PR #7, opened 2017-07-13
HZN-1073: Implement grafana datasource for the fault-management console

mvrueden:
This is the 2nd part of the issue.
It now uses the opennms.js client. 
The general structure was also re-defined and some tests are provided.
I think this can be the first version and the branch can be deleted afterwards. 

* JIRA: https://issues.opennms.org/browse/HZN-1073


j-white:
Can we use a released version of the opennms.js artifact from NPM instead of including it in the vendor directory?

src/datasources/fault-ds/client_delegate.js
```diff
@@ -0,0 +1,267 @@
...
+import _ from 'lodash';
+
+export class ClientDelegate {
+
+    constructor(settings, backendSrv, $q) {
+        this.type = settings.type;
+        this.url = settings.url;
+        this.name = settings.name;
+        this.backendSrv = backendSrv;
+        this.searchLimit = 1000;
```
j-white: Can we make this unlimited by default? We currently have no way of indicating the results have been truncated, or providing an order by clause that would help control which items are being cut.

PR #8, opened 2017-07-31
HELM-10: Allow users to edit journal and sticky notes.

j-white:
Here we add support for editing the journal and sticky memos via the alarm details modal.

The relevant calls are made via opennms.js, and the dependency is bumped to a version with those calls.

* JIRA: https://issues.opennms.org/browse/HELM-10


mvrueden:
I tested this and it seems to have some problems:
 * The alarm details are not populated. I only see the columns. It seems that `alarm` and `source` is undefined
 * I was able to add a jorunal/memo to an alarm (verified on the opennms side), but closing the dialog and reopen it, does not load the already existing data, so it is not possible to edit or delete or even see it.
 * The memo/journal input's background is white, which looks very weird in the dark theme. The input text color is grey-ish which makes it a bit weird. I think the color scheme should match with the grafana.
 * The save button does not have any padding. This is probably also true for the delete button, which I did not see (see above)

j-white:
Thanks for giving this a whirl.
1. Fixed.
1. Closing and re-opening the dialog won't reload the alarm, we continue to use the same alarm that is shown in the table panel. Instead you would need to refresh the dashboard. I opted for this behavior since it would be odd to see different contents in the details pane than those shown on the table panel.
1. Fixed.
1. I added some top padding, so that the buttons are no longer stuck against the text box.

That's okay, ticket states aren't relevant here.

mvrueden:
Looks good to me now. With the fixes made, the 2. point is no longer an issue. But now we have merge conflicts :-(

PR #9, opened 2017-08-02
HELM-7: create an RPM package from the Helm app

RangerRick:
This PR adds a script to generate an RPM package in `dist/packages/`.  If you want to test, it works fine with Homebrew's RPM.

### DIFF
--- a/makerpm.js
+++ b/makerpm.js
@@ -1,0 +1,99 @@
+#!/usr/bin/env node
+
+var os = require('os');
+var fs = require('fs-extra');
+var path = require('path');
+var rimraf = require('rimraf');
+var which = require('which');
+
+var validator = require('specit/lib/validator');
+var generate = require('specit/lib/generate');
+var clean = require('specit/lib/clean');
+
+var program = require('commander');
+
+var spawn = require('child_process').spawnSync;
+var cwd = process.cwd();
+
+var pkginfo = require('./package.json');
+
+var version = pkginfo.version;
+var release = 1;
+
+if (version.indexOf('-SNAPSHOT') > 0) {
+  version = version.replace('-SNAPSHOT', '');
+  release = '0';
+}
+
+try {
+  which.sync('rpmbuild');
+} catch (err) {
+  console.log('rpmbuild executable not found');
+  process.exit(1);
+}
+
+program
+  .version(pkginfo.version)
+  .option('-r --release <release>', 'Specify release number of package')
+  .option('-d --distribution <dist>', 'Specify target distribution.(.el5, .el7, .fc26, .fc22')
+  .parse(process.argv);
+
+pkginfo.version = version;
+pkginfo.release = release;
+if (program.release === undefined) {
+  program.release = release;
+}
+
+console.log('Generating RPM spec for ' + pkginfo.name + ' ' + pkginfo.version + '-' + program.release);
+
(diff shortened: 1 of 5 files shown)


RangerRick:
I've updated this to make the package not be 200MB of `node_modules`, and also a few other cleanup tweaks.  ;)

makerpm.js
```diff
@@ -0,0 +1,85 @@
...
+    console.log('Created ./%s', file);
+  });
+
+  var rpmbuilddir = path.join(os.tmpdir(), 'rpmbuild');
+  rimraf.sync(rpmbuilddir);
+  fs.mkdirSync(rpmbuilddir);
+  fs.mkdirSync(path.join(rpmbuilddir, 'SOURCES'));
+  fs.mkdirSync(path.join(rpmbuilddir, 'RPMS'));
+  fs.mkdirSync(path.join(rpmbuilddir, 'BUILD'));
+  fs.linkSync(path.join('SOURCES', pkginfo.name + '.tar.gz'), path.join(rpmbuilddir, 'SOURCES', pkginfo.name + '.tar.gz'));
```
RangerRick: Ah, yeah, shouldn't be significantly different.  Fixed.

makerpm.js
```diff
@@ -0,0 +1,85 @@
...
+    'rpmbuild',
+    [
+      '--define', '_topdir ' + path.join(os.tmpdir(), 'rpmbuild'),
+      '-ba',
+      'SPECS/opennms-helm-app.spec'
+    ],
+    {
+      stdio: ['inherit', 'inherit', 'inherit']
+    }
+  );
```
RangerRick: I added a `which` check, and then also exit(1) if the spawn fails.

j-white:
My version of node (v6.11.1) doesn't seem to like that `catch` statement you added:

```
$ ./makerpm.js 
/home/jesse/git/grafana-opennms-helm-app/makerpm.js:30
catch (err) {
^^^^^
SyntaxError: Unexpected token catch
```

RangerRick:
Haha yeah, turns out you're supposed to push *then* save in your editor... Should be fixed now.  Also fixed permissions to be owned by the `grafana` user.  (Should already be correct in the debian PR.)

j-white:
Cool. I had just noticed that permission problem too. Let me give this another whirl.

j-white:
I still get tons of these messages when installing:
```
warning: user opennms-helm-app does not exist - using root
warning: group opennms-helm-app does not exist - using root
```
Also, I don't think you meant to include the `opennms-helm-app` directory in this commit.

RangerRick:
Yeah, it looks like somewhere deep inside, `specit` is always setting the username to the package name.  I just changed the template to hardcode it to `grafana`.  Also yeah, deleted the helm app directory.  Perils of switching back and forth between the RPM and debian branches.  :)

j-white:
Permissions are good now, but I still have a problem with that catch statement:
```
$ ./makerpm.js 
/home/jesse/git/grafana-opennms-helm-app/makerpm.js:30
catch (err) {
^^^^^
SyntaxError: Unexpected token catch
```

RangerRick:
Hm, I'm not sure how, since line 30 doesn't look like that anymore; it's got the `}` at the beginning.

j-white:
Yep, my bad.

PR #10, opened 2017-08-02
HELM-8: implement debian package generation

RangerRick:
This is basically the debian equivalent of HELM-7, creating a .deb file in `dist/packages/` that depends on the upstream grafana package.


j-white:
I guess we need to rebase now that HELM-7 is merged.

PR #11, opened 2017-08-03
HELM-18: Add support for using template variables as query parameters

mvrueden:
* JIRA: https://issues.opennms.org/browse/HELM-18


j-white:
Works well.

PR #14, opened 2017-08-08
HELM-12: Add from and  to template variables

mvrueden:
* JIRA: https://issues.opennms.org/browse/HELM-12

### DIFF
--- a/src/datasources/fault-ds/datasource.js
+++ b/src/datasources/fault-ds/datasource.js
@@ -40,11 +40,14 @@ export class OpenNMSFMDatasource {
             if (clause.restriction instanceof API.NestedRestriction) {
                 self.substitute(clause.restriction.clauses, options);
             } else if (clause.restriction.value) {
-                const scopedVars = Object.assign({}, options.scopedVars, {
-                    "range_from":     {text: options.range.from.valueOf(),   value: options.range.from.valueOf()},
-                    "range_to":  {text: options.range.to.valueOf(), value: options.range.to.valueOf()},
-                });
-                clause.restriction.value = self.templateSrv.replace(clause.restriction.value, scopedVars);
+                // Range must be of type date, otherwise it is not parseable by the OpenNMS client
+                if (clause.restriction.value === '$range_from' || clause.restriction.value === "[[range_from]]") {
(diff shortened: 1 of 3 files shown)


src/datasources/fault-ds/datasource.js
```diff
@@ -40,9 +40,8 @@ export class OpenNMSFMDatasource {
             if (clause.restriction instanceof API.NestedRestriction) {
                 self.substitute(clause.restriction.clauses, options);
             } else if (clause.restriction.value) {
-                // TODO MVR: This is a hint on how to probably best implement HELM-12
-                // clause.restriction.value = clause.restriction.value.replace(/\$timeFrom/g, options.range.from.valueOf());
-                // clause.restriction.value = clause.restriction.value.replace(/\$timeTo/g, options.range.to.valueOf());
+                clause.restriction.value = clause.restriction.value.replace(/\$range_from/g, options.range.from.valueOf());
+                clause.restriction.value = clause.restriction.value.replace(/\$range_to/g, options.range.to.valueOf());
                 clause.restriction.value = self.templateSrv.replace(clause.restriction.value, options.scopedVars);
```
j-white: Should we not add the `range_from` and `range_to` to the list of `scopedVars`, rather than performing the token replacement ourselves?

mvrueden: Yep. Done.

mvrueden:
Somehow this does not work locally for me, even with the latest date enhancements to the opennms.js library. I have to investigate.

mvrueden:
The problem is, that the grafana build in template variable service is doing string replacement. That means, that a `$range_from` or `$range_to` variable is converted using the String.replace(...) method. The original proposed solution would convert `$range_from`/`$range_to` to a timestamp. That would work, with the `templateSrv`, however the opennms.js library does not know that. That functionality could be added.

The latest proposed solution is setting the interval date directly instead converting it to a number. That way no change must be made to the opennms.js client. However a partial string replacement no longer works.

I think both implementations have their pros and cons and I am not in favor of any solution. Just some input on the issue at hand.

j-white:
I think the current solution is fine.

Can we add a test case for this?

PR #18, opened 2017-08-14
HELM-35: Defer null value handling to opennms.js client

mvrueden:
* JIRA: https://issues.opennms.org/browse/HELM-35

Must be merged after PR https://github.com/OpenNMS/opennms-js/pull/14

### DIFF
--- a/src/datasources/fault-ds/mapping/ValueMapping.js
+++ b/src/datasources/fault-ds/mapping/ValueMapping.js
@@ -1,34 +1,10 @@
 
 export class ValueMapping {
     getApiValue(internalAttribute, value) {
-        // TODO MVR value mapping should be moved to the opennms-js implementation
-        // if (internalAttribute === 'alarmAckTime') {
-        //     if (value === 'null') {
-        //         value = new Date(0);
-        //     }
-        //
-        //     // yyyy-MM-ddTHH:mm:ss.sssZ
-        //     // 2017-06-08T10:17:17.173+0200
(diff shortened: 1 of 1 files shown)


src/datasources/fault-ds/mapping/ValueMapping.js
```diff
@@ -1,34 +1,10 @@
 
 export class ValueMapping {
```
j-white: Is it worth even keeping this class, if all it does is return the given `value`?

mvrueden: Let's keep it for now, until we know that we really don't need it. My plan is to clean up the code when we are ready for v1. I just don't want to delete it now and then re-create it later (-:

j-white:
Can we add a note in the query editor to indicate that 'null' is a valid value for the date fields?

j-white:
Also, should be bump the version of opennms-js that we depend on here? Since https://github.com/OpenNMS/opennms-js/pull/14 has been merged.

mvrueden:
The explanation was added directly in develop with the following commit: https://github.com/OpenNMS/grafana-opennms-helm-app/commit/e4f349eadecc046b52b8c35a5307f419214d8a94

The opennms.js version was bumped directly in develop with the following commit: https://github.com/OpenNMS/grafana-opennms-helm-app/commit/aad57bd8881c992a020bd677f67d09bba39fdbef

PR #26, opened 2017-08-21
HELM-33: Fix de-serialization issue with target filter

mvrueden:
* JIRA: https://issues.opennms.org/browse/HELM-33


j-white:
Fixed worked for me, I was no longer able to reproduce the issue.

PR #28, opened 2017-08-22
HELM-26: Flatten all alarm parameters

mvrueden:
* JIRA: https://issues.opennms.org/browse/HELM-26

Missing attributes are `Event Label` and `Ticket URL`, but they are not yet exposed via the rest endpoints.


j-white:
Looks good.

Event label is tracked in : https://issues.opennms.org/browse/HELM-13
Ticket URL is tracked in: https://issues.opennms.org/browse/HELM-31

PR #34, opened 2017-08-24
HELM-38: rename "opennms-helm-app" to "opennms-helm"

RangerRick:
This PR renames "opennms-helm-app" to "opennms-helm" everywhere it is found.  It also fixes URLs to point to the renamed github repo, rather than "grafana-opennms-helm-app".


j-white:
```
$ find . -type f | xargs grep -i 'helm-app' | grep -v '.git' | wc -l
0
```

PR #35, opened 2017-08-24
HELM-48: Rename data sources for clarity.

j-white:
JIRA: https://issues.opennms.org/browse/HELM-48


RangerRick:
Pretty obvious change and it passed tests.  :D

PR #36, opened 2017-08-24
add a note about installing node.js first for debian/rpm installs

RangerRick:
Normally for CentOS 7 installation, you would just enable the `epel` repo and get Node.js there, but since RHEL7.4 is about to come out, EPEL has removed some of Node's dependencies and installation is broken.

This patch adds a note about installing Node.js through the official upstream's recommendation for package installs so that we don't have to worry about broken dependencies.


RangerRick:
Never mind, this was only necessary because the RPMs incorrectly had a runtime dependency on `nodejs`.

PR #40, opened 2017-08-30
HELM-51: Redefine defaults for Alarm Table Panel

mvrueden:
* JIRA: https://issues.opennms.org/browse/HELM-51


j-white:
Merged manually with some other tweaks.

PR #47, opened 2018-02-12
HELM-79: add support for custom context menu actions

RangerRick:
This PR adds support for adding custom actions to context menus.  Currently it is only supported in the alarm table, but the configuration and code are done in such a way that they could be reused elsewhere.


j-white:
Is there a way to access event parameters by name?

I tried triggering an alarm with:
```
$OPENNMS_HOME/bin/send-event.pl uei.opennms.org/alarms/trigger \                                         
  -p 'node n33' \                                   
  -p 'ip 0.0.0.0' \                                 
  -p 'service HTTPS_POOLs' \                        
  -p 'type ENV' \                                   
  -p 'cat1 ENVIRONMENTAL' \                         
  -p 'cat2 HUMIDITY' \                              
   -x 7 \                                           
   -n 4
```

And using the following URL:
```
http://127.0.0.1:8980/test?cat=$parameters[cat1]&cat=$parameters[cat2]
```



The action is not displayed. Is there

j-white:
When a variable is replaced, is it URL encoded? i.e. What if my event parameter value contains `&`

RangerRick:
Hmm, good points. I'll work on both of those.

RangerRick:
@j-white alright, I've done both of those changes

pbrane:
Nice work, Ben!

src/panels/alarm-table/action_mgr.js
```diff
@@ -75,6 +77,23 @@ export class ActionMgr {
...
+
+    if (self.rows.length === 1 && self.appConfig.actions && self.appConfig.actions.length > 0) {
+      for (let action of self.appConfig.actions) {
+        if (!action.label || !action.url || action.label.trim().length === 0 || action.url.trim().length === 0) {
+          console.warn('invalid label or URL:',action);
+          continue;
+        }
+        const a = new CustomAction(action);
+        const row = self.rows[0];
+        console.log('row=',row);
```
j-white: Can we remove/comment this out?

src/lib/custom_action.js
```diff
@@ -0,0 +1,149 @@
...
+
+  validate(model) {
+    const interpolated = this.url;
+    let passed = true;
+    if (model) {
+      for (const key of Object.keys(model)) {
+        const match = getMatch(interpolated, key);
+        const value = getValueForMatch(match, key, model);
+        if (value === failed) {
+          console.warn('Variable $' + key + ' was found in the model object, but no value was found.', model);
```
j-white: This can happen as part of "normal" usage if we refer to an event parameter that's only on *some* alarms. Is it worth always logging a warning?

PR #48, opened 2018-03-16
refactor docs to list requirements in one place

RangerRick:


docs/installation/requirements.adoc
```diff
@@ -0,0 +1,7 @@
+= Requirements
+
+* http://docs.grafana.org/installation[Grafana]
+The required Grafana-Server version is *4.4.1* up to *< 6*.
+* https://yarnpkg.com/en/docs/install[yarn]
```
j-white: We don't actually require nodejs or yarn for the runtime - only for the build.

RangerRick: ah, good point

PR #50, opened 2018-04-02
HELM-86: add support for configurable timeouts

RangerRick:
This PR builds on [HELM-78](https://github.com/OpenNMS/opennms-helm/pull/49) to add configurable timeout support to the Helm datasources.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "babel-plugin-transform-es2015-modules-systemjs": "^6.24.1",
     "babel-preset-es2015": "^6.24.1",
     "lodash": "~4.17.4",
-    "opennms": "^1.1.1",
+    "opennms": "^1.2.0",
     "q": "^1.5.0"
   },
   "homepage": "https://github.com/OpenNMS/opennms-helm",
--- a/src/components/timeout.html
+++ b/src/components/timeout.html
@@ -1,0 +1,13 @@
+<div class="section gf-form-group">
+    <h3 class="page-heading">
+        Miscellaneous HTTP Settings
+    </h3>
+    <div class="gf-form-group">
+        <div class="gf-form-inline">
+            <div class="gf-form">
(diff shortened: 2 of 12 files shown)


j-white:
If I point the datasource to use a host that will timeout i.e. `https://34.194.50.139:8443/opennms`, set the timeout to `30`, and click `Save & Test`. The calls fails after `10` seconds with:
```
{"err":{"data":null,"status":-1,"config":{"method":"GET","transformRequest":[null],"transformResponse":[],"jsonpCallbackParam":"callback","auth":{},"timeout":10000,"headers":{"accept":"application/json","X-Requested-With":"XMLHttpRequest","X-Grafana-Org-Id":1,"X-Grafana-NoCache":"true"},"parameters":{},"server":{"id":"93e3a27b-e62a-45b5-b66a-7b5b5744bb3a","name":"minion-dev (Flow)","url":"/api/datasources/proxy/10","auth":{}},"url":"api/datasources/proxy/10/rest/info","retry":0},"statusText":"","xhrStatus":"abort"},"cancelled":true}
```

I see similar behavior when using the datasource in the panels - the requests continue to timeout after 10 seconds regardless of the value entered in the datasource settings.

RangerRick:
Hrm, OK, I'll take another look.  When I tried it before it was working, but I did refactor a little since then.

RangerRick:
It was making it to the `client.http.options` but those were being ignored by `getClientWithMetadata`.  I've changed the code to pass a timeout to `getClientWithMetadata` but for the long term I've made [a PR against OpenNMS.js](https://github.com/OpenNMS/opennms-js/pull/25) to deprecate the `timeout` property altogether and favor the one in the `options` property.

j-white:
That seems to have done the trick. Timeouts are being applied properly now.

I did run into this issue when testing the PM datasource, but we can handle that separately: https://issues.opennms.org/browse/HELM-88

PR #51, opened 2018-04-04
HELM-87: implement fallback-attribute in grafana perf-ds

RangerRick:
Pretty straightforward, this PR implements support for `fallback-attribute` in the performance datasource.  I confirmed it by typo-ing the attribute and then providing a valid value (`ifInOctets`) for the `fallback-attribute` and I still get graphs, so it seems to be working as expected.  :)

### DIFF
--- a/src/datasources/perf-ds/datasource.js
+++ b/src/datasources/perf-ds/datasource.js
@@ -189,6 +189,7 @@ export class OpenNMSDatasource {
         var source = {
           "aggregation": target.aggregation,
           "attribute": target.attribute,
+          "fallback-attribute": target.fallbackAttribute,
           "label": label,
           "resourceId": target.resourceId,
           "nodeId": target.nodeId, // temporary attribute used for interpolation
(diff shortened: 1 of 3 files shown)


src/datasources/perf-ds/partials/query.editor.html
```diff
@@ -87,6 +87,21 @@
                     >
                 </div>
             </div>
+            <div class="gf-form-inline">
+                <div class="gf-form max-width-25">
+                    <label class="gf-form-label query-keyword width-10">
+                        Fallback Attribute <i class="fa fa-tag"></i>
```
j-white: It'd be nice to be able to use the  "Attribute selection modal" here to be able to lookup the value. Otherwise you need to either know the attribute name, or use the other modal, copy the value and dismiss it so that it doesn't overwrite the existing attribute value.

src/datasources/perf-ds/datasource.js
```diff
@@ -189,6 +189,7 @@ export class OpenNMSDatasource {
         var source = {
           "aggregation": target.aggregation,
           "attribute": target.attribute,
+          "fallback-attribute": target.fallbackAttribute,
```
j-white: Can we omit this from the query if there is no fallbackAttribute defined? (similar to what we do with the 'subattribute')

j-white:
👍

PR #52, opened 2018-04-05
HELM-88: fix PM datasource failure handling

RangerRick:
This PR fixes the PM datasource to properly return a rejection when the request fails.  It also tweaks the failure object so it provides a bit more data in the inspector.

### DIFF
--- a/src/datasources/perf-ds/datasource.js
+++ b/src/datasources/perf-ds/datasource.js
@@ -55,12 +55,11 @@ export class OpenNMSDatasource {
       });
     } else {
       // There are no sources listed, use an empty set of measurements
-      request = this.$q.defer();
-      request.resolve({measurements: []});
+      request = this.$q.resolve({measurements: []});
     }
 
     // Convert the results to the expected format
-    return this.$q.when(request).then(function (response) {
+    return request.then((response) => {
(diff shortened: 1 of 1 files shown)


j-white:
On timeout, I now get a message in the console:
   `Panel request cancelled {err: {…}, cancelled: true, message: "An unexpected error occurred."}`
but the Graph panel keeps spinning.

Did you observe different behavior?

RangerRick:
Hm, I do observe different behavior.

It doesn't spin for me, and it doesn't show as cancelled.  Instead, when I stop my OpenNMS I get:

```
Panel data error: {data: "", status: 502, headers: ƒ, config: {…}, statusText: "Bad Gateway", …}
```

What browser/OpenNMS version/resources/attributes?  Also, do you see any data in the inspector?

RangerRick:
Ah yeah, I get you.  I'll see if I can come up with something listening on 8980 that never responds and try to reproduce your issue.

RangerRick:
Aha, if the `cancelled` property is on the object, it never finishes.  I've pushed an updated version that seems to handle it right for me.  Can you confirm?

j-white:
Works much better now.

PR #58, opened 2018-07-03
HELM-97: Fix for "Template variable were only replaced once"

michael-batz:
The String.replace() function only replaces the first occurence of a found element, when used with a substring as parameter. I changed the parameter to a RegExp to replace all occurences of template variables.

### DIFF
--- a/src/datasources/perf-ds/interpolate.js
+++ b/src/datasources/perf-ds/interpolate.js
@@ -60,7 +60,7 @@ function defaultReplace(value, variables) {
   }
   var interpolatedValue = value;
   _.each(variables, function (variable) {
-    interpolatedValue = interpolatedValue.replace("$" + variable.name, variable.value);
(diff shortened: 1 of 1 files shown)


src/datasources/perf-ds/interpolate.js
```diff
@@ -60,7 +60,7 @@ function defaultReplace(value, variables) {
   }
   var interpolatedValue = value;
   _.each(variables, function (variable) {
-    interpolatedValue = interpolatedValue.replace("$" + variable.name, variable.value);
+    interpolatedValue = interpolatedValue.replace(new RegExp("\\$" + variable.name, "g"), variable.value);
```
j-white: What if the variable name contains reserved regex characters? We need to escape those as well for this technique to work reliably.

michael-batz:
@j-white you are right. I escaped the special characters.

j-white:
Thanks @michael-batz!

PR #59, opened 2018-07-04
HZN-1328 Add an additional tab to Alarm display related Alarms

smith-opennms:
This PR adds an additional tab to the Alarm Detail view if the Alarm being viewed has Related Alarms.

JIRA: https://issues.opennms.org/browse/HZN-1328


src/datasources/fault-ds/datasource.js
```diff
@@ -192,6 +192,7 @@ export class OpenNMSFMDatasource {
             "Suppressed Time", "Suppressed Until", "Suppressed By", "IP Address",
             "First Event Time", "Last Event ID", "Last Event Time", "Last Event Source",
             "Last Event Creation Time", "Last Event Severity", "Last Event Label", "Last Event Location",
+            "RelatedAlarms",
```
j-white: I think we can skip adding this to column to the row - the related alarms don't render well in a table cell.

src/panels/alarm-table/alarm_details.html
```diff
@@ -90,5 +90,25 @@ <h3>Journal Memo</h3>
                 </tr>
             </table>
         </div>
+
+        <div ng-if="editor.index > 1  && tabs[editor.index] == 'Related Alarms' && alarm.relatedAlarms && alarm.relatedAlarms.length > 0">
+            <table  class="table-panel-table dense-table-panel-table alarm-table">
+                <thead>
+                    <th> </th>
+                    <th>ReducktionKey</th>
```
j-white: Can we change `ReducktionKey` to `Reduction Key`?

src/datasources/fault-ds/datasource.js
smith-opennms: done.

PR #60, opened 2018-07-07
HELM-100: allow overriding time intervals and max datapoints

brynjare:
This commit adds support for setting a global time interval for the PM datasource and optionally overriding it per query. Max datapoints can also be specified per query.

Allowing a global default is useful when the minimum_step setting in OpenNMS is lower than what you'd want to use for most graphs. (For example when polling certain services every 60 seconds, but everything else at a larger interval.)

One new configuration option for the PM datasource:
intervals_global

Two new graph query options:
intervals_query

### DIFF
--- a/src/datasources/perf-ds/datasource.js
+++ b/src/datasources/perf-ds/datasource.js
@@ -10,6 +10,7 @@ export class OpenNMSDatasource {
     this.name = instanceSettings.name;
     this.basicAuth = instanceSettings.basicAuth;
     this.withCredentials = instanceSettings.withCredentials;
+    this.interval = (instanceSettings.jsonData || {}).timeInterval;
 
     if (instanceSettings.jsonData && instanceSettings.jsonData.timeout) {
         this.timeout = parseInt(instanceSettings.jsonData.timeout,10) * 1000;
@@ -191,6 +192,7 @@ export class OpenNMSDatasource {
       start = options.range.from.valueOf(),
       end = options.range.to.valueOf(),
       step = Math.floor((end - start) / options.maxDataPoints);
+      step = (step < options.intervalMs) ? options.intervalMs : step;
 
     var query = {
       "start": start,
--- a/src/datasources/perf-ds/plugin.json
+++ b/src/datasources/perf-ds/plugin.json
@@ -10,6 +10,11 @@
   "metrics": true,
   "annotations": false,
 
+  "queryOptions": {
+    "minInterval": true,
+    "maxDataPoints": true
+  },
+
   "info": {
     "logos": {
       "small": "img/pm-ds.svg",
--- a/src/datasources/perf-ds/partials/config.html
+++ b/src/datasources/perf-ds/partials/config.html
@@ -2,3 +2,19 @@
 </datasource-http-settings>
 <onms-timeout-settings current="ctrl.current">
 </onms-timeout-settings>
+<h3 class="page-heading">OpenNMS PM Datasource Details</h3>
+<div class="section gf-form-group">
+    <div class="gf-form-inline">
+        <div class="gf-form">
+            <span class="gf-form-label">Min time interval</span>
+            <input type="text" class="gf-form-input width-6" ng-model="ctrl.current.jsonData.timeInterval" spellcheck='false' placeholder="10s"></input>
+            <info-popover mode="right-absolute">
+                A lower limit for the auto group by time interval. Recommended to be set to write frequency,
+                for example <code>1m</code> if your data is written every minute.
+
+                When using Newts as the Time Series Strategy, org.opennms.newts.query.minimum_step may end up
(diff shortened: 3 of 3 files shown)


src/datasources/perf-ds/plugin.json
```diff
@@ -10,6 +10,11 @@
   "metrics": true,
   "annotations": false,
 
+  "queryOptions": {
```
j-white: What does this do?

src/datasources/perf-ds/datasource.js
```diff
@@ -10,6 +10,7 @@ export class OpenNMSDatasource {
     this.name = instanceSettings.name;
     this.basicAuth = instanceSettings.basicAuth;
     this.withCredentials = instanceSettings.withCredentials;
+    this.interval = (instanceSettings.jsonData || {}).timeInterval;
```
j-white: The variable `interval` is set, but doesn't appear to be used anywhere?

brynjare: 'interval' isn't used directly, but referenced through 'intervalMs' instead. The calculation is [handled internally by Grafana](https://github.com/grafana/grafana/blob/25bcdbcab1b7e2329e6b147e4c505b29d00fb7a3/public/app/features/panel/metrics_panel_ctrl.ts#L157-L170).

They are also available to the users as  [$__interval and $__interval_ms](http://docs.grafana.org/reference/templating/#the-interval-variable) respectively.

src/datasources/perf-ds/plugin.json
brynjare: It makes Grafana automatically include the two "graph query options" from the screenshot in the pull request. It pretty much just makes Grafana know that the datasource supports overriding these settings.

[torkelo's comment](https://github.com/grafana/grafana/issues/9411#issuecomment-333934429) is the closest thing I could find to "documentation" of queryOptions.

[This is the code](https://github.com/grafana/grafana/blob/3a1f52d8a2fdd1fd3b9bdbee05a48925bf8580f1/public/app/features/panel/partials/metrics_tab.html#L36-L63) that handles the user interaction / presentation.

'maxDataPoints' isn't really necessary, I suppose, but I included it since it's used by 'maxrows' in the API calls sent to OpenNMS.

roskens:
I tried this out against Grafana 5.1.4, and this works as expected.

I setup a dashboard with a table panel, and the times returned for all metrics were adjusted based on the options specified for the OpenNMS Data Source. So instead of seeing one every 10 minutes, I was seeing results for a whole hour like I wanted.

j-white:
Thanks for the feedback @roskens.

src/datasources/perf-ds/plugin.json
j-white: Thanks for the reference, wasn't aware of that one.

src/datasources/perf-ds/datasource.js
j-white: Right. I guess what I'm getting at is do we need this line?

brynjare: Yes, this is how we access the default value configured in the [datasource settings](https://github.com/OpenNMS/opennms-helm/pull/60/files#diff-ee658e695e923a4837d086d30aeb078bR10).

The default value overrides Grafana's own assumption about a suitable time interval, and then [calculateInterval()](https://github.com/grafana/grafana/blob/25bcdbcab1b7e2329e6b147e4c505b29d00fb7a3/public/app/features/panel/metrics_panel_ctrl.ts#L157-L170) *optionally* overrides the default value as well, with whatever the user specifies in the query options for a specific panel.

Without this line, Grafana will just make a guess on what it thinks the time interval should be, based on the selected time range. When looking at the "last 30 minutes", it figures 2 seconds is a suitable interval, which `org.opennms.newts.query.minimum_step` will then override.

`web.log` used to be filled with benign, but annoying, warnings like this before I added the default configuration:
```
2018-07-12 22:40:05,453 WARN  [qtp1059610547-58601] o.o.n.m.i.NewtsFetchStrategy: Requested step size 22500 is too small. Using 60000.
2018-07-12 22:42:44,862 WARN  [qtp1059610547-58601] o.o.n.m.i.NewtsFetchStrategy: Requested step size 22500 is too small. Using 60000.
2018-07-12 22:46:53,097 WARN  [qtp1059610547-59227] o.o.n.m.i.NewtsFetchStrategy: Requested step size 22500 is too small. Using 60000.
2018-07-12 22:47:10,444 WARN  [qtp1059610547-58453] o.o.n.m.i.NewtsFetchStrategy: Requested step size 5000 is too small. Using 60000.
```

j-white: Got it. I had missed the fact that the `calculateInterval` call retrieved the interval via `this.panel.interval`, which references the variable in question.

PR #65, opened 2018-09-05
OCE - ReST updates

smith-opennms:
This PR allows the helm Alrm tables to be built : 
* filtering on whether an Alarm is in a situation or not
* by the number of nodes the alarm or situation affects (affectedNodeCount) and 
* by the number of correlated alarms in the situation (situationAlarmCount).

The latter 2 are displayable as columns.

### DIFF
--- a/src/datasources/fault-ds/datasource.js
+++ b/src/datasources/fault-ds/datasource.js
@@ -197,7 +197,7 @@ export class OpenNMSFMDatasource {
             "Last Event Creation Time", "Last Event Severity", "Last Event Label", "Last Event Location",
             "Sticky ID", "Sticky Note", "Sticky Author", "Sticky Update Time", "Sticky Creation Time",
             "Journal ID", "Journal Note", "Journal Author", "Journal Update Time", "Journal Creation Time",
-            "Is Situation", "Is in Situation", "Situation Alarm Count", "Affected Node Count",
+            "Is Situation", "Situation Alarm Count", "Affected Node Count",
             "Data Source"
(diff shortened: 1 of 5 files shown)


src/datasources/fault-ds/datasource.js
```diff
@@ -5,9 +5,9 @@ import {Mapping} from './Mapping';
 import _ from 'lodash';
 
 const FeaturedAttributes = [
-    "alarmAckTime", "category", "ipAddress",
-    "isSituation", "location", "node.label", "reductionKey",
-    "service", "severity", "uei"
+    "affectednodeCount", "alarmAckTime", "category", "ipAddress",
```
j-white: Are these case sensitive? `affectednodeCount` vs `affectedNodeCount`

src/panels/alarm-table/alarm_details.html
```diff
@@ -106,7 +106,7 @@ <h3>Journal Memo</h3>
                                 <i class="fa fa-external-link"></i>
                             </a>
                         </td>
-                        <td severity>{{summary.reductionKey}}</td>
+                        <td severity>{{summary.label}}</td>
                         <td severity ng-bind-html="summary.description"></td>
```
j-white: The table header was changed from description to log message, but we still render the description.

src/datasources/fault-ds/datasource.js
smith-opennms: yes - case sensitive - will fix.

PR #67, opened 2018-09-14
HELM-104: Create automated build for Docker Images

indigo423:
Create a workflow to build and publish a Docker image to DockerHub. The workflow provides the following steps on CircleCi:

 1. Build and test from source
 2. Create a tarball of the build artifact
 3. Create a Docker Image with installing from the tarball artifact
 4. Only publish on Dockerhub when a git tag as version number is set

Steps 1-3 are triggered for every branch including pushing a git tag. Publish is only triggered when a git tag is pushed. The version number from the git tag is used for the Docker Image and during the plugin build.

For any other branches, the tarball and a Docker Image is created but stored as artifact on CircleCi under the specific build number and can be downloaded from the artefact section of the build job. The dependency for Grafana version can be controlled in `grafana.version`. With this commit it is set to floating stable release tagged as `grafana/grafana:latest`.

To control the DockerHub publish job, the following environment variables for the build are introduced and are set in [CircleCi project](https://circleci.com/gh/OpenNMS/workflows/opennms-helm) settings:

* `DOCKERHUB_LOGIN`: Login on DockerHub
* `DOCKERHUB_PASS`: Password on DockerHub
* `DOCKERHUB_PROJECT_USER`: Name of the user or organisation where the image is pushed to
* `DOCKERHUB_PROJECT_NAME`: Name of the application

JIRA: https://issues.opennms.org/browse/HELM-104


j-white:
Can we use a fixed version number instead of `latest` as the default for the Grafana base images? While we do generally want to be on the latest, we want to make sure the build are reproducible and don't break on us unexpectedly.

Also, can we make it so that builds from the `master` branch are published as well under either `latest` or `bleeding` tags. We're going to start depending on these artifacts for end-to-end testing, so having up-to-date images with the latest code would be a big help.

Aside from those comments, everything looks great.

indigo423:
> Also, can we make it so that builds from the `master` branch are published as well under either `latest` or `bleeding` tags. We're going to start depending on these artifacts for end-to-end testing, so having up-to-date images with the latest code would be a big help.

I haven't fully understood the workflow what triggers a release and develop / master and tags are used. Which is the reason I've picked the Tag as a controllable trigger for a release and getting a reasonable version number. 

I need help with the following questions:

* What is the source branch for a stable versioned release with a floating tag `latest`? What is the source for the version number (git tag/somewhere in code)? What should be the trigger for a push to DockerHub? A commit and/or a Git Tag push?
* What is the source branch for snapshot or bleeding release? What is the source for the version number (git tag/somewhere in code)? What should be the trigger for a release on DockerHub? For bleeding it seems to be legit to use a commit to ${branch}.

j-white:
> Sure no problem, I didn't know which version so I've picked latest. Just tell me which version I should pick instead.

5.2.4

j-white:
> It seems like we don't follow a [GitFlow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) pattern here. I need help with the following questions:
> 
> * What is the source branch for a stable versioned release with a floating tag `latest`? What is the source for the version number (git tag/somewhere in code)? What should be the trigger for a push to DockerHub? A commit and/or a Git Tag push?

Normally, the `latest` tag points to the latest stable release. When we release, we create a new tag off of the `master` branch. So `latest` should typically point to the last tagged release.

> * What is the source branch for snapshot or bleeding release? What is the source for the version number (git tag/somewhere in code)? What should be the trigger for a release on DockerHub? For bleeding it seems to be legit to use a commit to ${branch}.

Bleeding should run off the of `master` branch. Every commit to `master` should push an updated image.

In terms of workflow, developers currently target the `develop` branch with their code changes, and we have a job (currently running in Bamboo) that merges `develop` to `master` and includes the updated artifacts as part of the push. `master` includes a copy of the built `dist` directory, whereas `develop` does not.

indigo423:
The only way to get it to work was with two dedicated docker publish jobs, one for bleeding which is trigged on any commit on master and one for a release which is triggered by a git push --tag.

IMHO, the way we use the master branch to deliver the `dist` directory as an artefact smells :).

@j-white @RangerRick review if you can :)

PR #68, opened 2018-10-05
HELM-62: make "Severity" a column type

RangerRick:
This PR changes the `TableRenderer` to treat "severity" as a first-class type, like number or date columns.

* bump version to 3.0.0 because of breaking config change
* remove config option for "Severity icon"
* add column type "severity"
* add support for configuring severity column type (icon, label, upper-case label)
* add support for sorting the severity column
* add a mouseover to the severity icon

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ public/css/*.min.css
 
 # Editor junk
 *.sublime-workspace
+.vscode/
 *.swp
 .idea/
 *.iml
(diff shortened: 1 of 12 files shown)


j-white:
I ran into this when trying to create a new alarm table panel:
image

j-white:
Worked well for me with the update.

I think we may still have a problem with upgrades though, it's pretty hard to restore the severity column to the beginning of the column list without https://issues.opennms.org/browse/HELM-63. Right now users would need to delete all of the existing columns, then add the severity column, and the other columns back in order.

RangerRick:
Hrm, yeah.  Seems like since HELM-63 is a feature request anyways, makes sense to do it now, too.

PR #69, opened 2018-10-05
HELM-99: make severity color coding column or row

RangerRick:
This PR makes the color-coding for severity configurable to either be `Row` (the entire row, like the current impl when checked), `Column` (just any column with `type='severity'`), or `Off` (like the current impl, when unchecked).

This PR relies on changes made in https://github.com/OpenNMS/opennms-helm/pull/68


j-white:
Rebased over develop now that HELM-62 was merged.

PR #70, opened 2018-10-09
HELM-63: allow reordering of elements

RangerRick:
Add support for reordering columns in the alarm table config.


smith-opennms:
LGTM

j-white:
The header of the table get immediately updated after the columns have been re-ordered, but the body of the table isn't updated accordingly until the panel is refresh.

Is it feasible to make the body of the table update immediately too?

RangerRick:
@j-white hrm, good question, I'll see if I can trigger a re-render

RangerRick:
Yup, just needed to re-render the panel after the columns changed and it fixed it.

PR #72, opened 2018-10-16
HELM-73: make $node template variable work in alarm queries

RangerRick:
This PR fixes the `fault-ds` so it turns a `node=foreignSource:foreignId` criteria from a template variable into a separate set of criterias for foreign source and foreign ID.

Note: it does _not_ solve the (probably common) issue of wanting to multi-select nodes.  That will require additional work.

### DIFF
--- a/src/datasources/fault-ds/datasource.js
+++ b/src/datasources/fault-ds/datasource.js
@@ -74,16 +74,29 @@ export class OpenNMSFMDatasource {
       const self = this;
       _.each(clauses, clause => {
         if (clause.restriction) {
-            if (clause.restriction instanceof API.NestedRestriction) {
-                self.substitute(clause.restriction.clauses, options);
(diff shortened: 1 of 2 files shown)


j-white:
Looks good - haven't tried it yet though. Can we add unit test coverage for this?

RangerRick:
This is when I realize I should put the filtering in the `datasource.buildQuery` instead because it's gonna be a lot easier to test.  ;)

j-white:
A couple things:
1) 'node' isn't a field that shows up in the available list of fields, so we may want to add it there. Might be worth adding some documentation to explain what it's used for.
2) This doesn't work if the referenced node doesn't have a FS:FID - in this case the nodeFilter() query will just set the value as the node id

j-white:
👍  (I added a note to HELM-5 w.r.t. documenting the 'node' field.)

PR #79, opened 2018-11-29
HELM-123: make drag&drop work in IE and Firefox

brynjare:
* JIRA: https://issues.opennms.org/browse/HELM-123

This seems to fix drag&drop reordering in IE 11 and Firefox 63, while still working in Chrome.

(There seemed to be quite a few different ways to get the same element, however, so I'm not sure if this is the appropriate way. Feel free to change / improve it before merging.)


RangerRick:
Confirmed in FireFox, Chrome & Safari... Don't have IE handy but seems Good Enough (TM) to me.

PR #81, opened 2018-12-09
NMS-8942: Preserve order in queries

fooker:
https://issues.opennms.org/browse/NMS-8942

### DIFF
--- a/src/spec/perf_ds_datasource_spec.js
+++ b/src/spec/perf_ds_datasource_spec.js
@@ -302,20 +302,22 @@ describe('OpenNMSPMDatasource', function () {
     it('should preserve multiple labels (reverse)', function () {
         let options = {
             range: {from: 'now-1h', to: 'now'},
-            targets: [{
-                type: "attribute",
-                nodeId: '1',
-                resourceId: 'nodeSnmp[]',
-                attribute: 'loadavg5',
-                aggregation: 'AVERAGE'
-            },
-            {
-                type: "attribute",
-                nodeId: '1',
-                resourceId: 'nodeSnmp[]',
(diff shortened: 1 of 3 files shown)


fooker:
The query is not re-executed if one changes the order. Therefore you have to click in one of the fields in the query editor to see the changes. I'm not sure if this is by design or not...

fooker:
Looks like the missing re-execution on order changes is an upstream issue.

fooker:
Reverted strange renaming from `order` to `labels`

fooker:
Tests added and pulled latest develop

src/spec/perf_ds_datasource_spec.js
```diff
@@ -251,4 +251,79 @@ describe('OpenNMSPMDatasource', function () {
...
+                resourceId: 'nodeSnmp[]',
+                attribute: 'loadavg1',
+                aggregation: 'AVERAGE'
+            }],
+            interval: '1s'
+        };
+
+        let labels = ctx.ds.buildQuery(options).labels;
+
+        expect(labels.length).to.equal(2);
```
j-white: This verifies that an array of labels are returned by the `buildQuery()` method. Can we also validate that the resulting series are actually reordered to match this? (The code is pretty simple, but this will help make sure it doesn't get inadvertently broken down the road.)

fooker: Added

RangerRick:
Yeah as far as I can tell this is completely broken if you use template variables.  I can't quite figure out how to deal with it yet so I'm gonna revert it for now until @fooker can look into it more.

PR #84, opened 2019-01-17
HELM-84: support formatting of values from `nodeFilter` and `nodeResources`

RangerRick:
This PR adds support for formatting labels based on metadata returned by Meridian 24 and later. (The backend changes are in [this pull request](https://github.com/OpenNMS/opennms/pull/2318).)

It adds support for the following "functions" to be used in the label field:

* `nodeToLabel(<nodeId|foreign-source:foreign-id>)`: the associated node's label
* `resourceToLabel(<nodeCriteria|resourceId>[, partialResourceId])`: the resource's label
* `resourceToName(<nodeCriteria|resourceId>[, partialResourceId])`: the resource's name
* `resourceToInterface(<nodeCriteria|resourceId>[, partialResourceId])`: the interface name

The resource functions can accept either a single argument containing a complete resource ID (`node[situation-test:nodeb].interfaceSnmp[en0-000123456789]`), or two arguments of a node criteria and partial resource ID (`foo:bar, interfaceSnmp[en0-000123456789]` or `1, interfaceSnmp[en0-000123456789]`).

PR #85, opened 2019-01-17
Initial documentation for the performance datasource.

j-white:
This isn't exhaustive, but it's a start...


RangerRick:
I guess it's implicit in me adding to it, but 👍 from me on this.  Should we go ahead and merge it after #84 ?

PR #87, opened 2019-01-29
HELM-129: Use the Measurements API in relaxed mode

j-white:
JIRA: https://issues.opennms.org/browse/HELM-129

Here we set `relaxed=true` by default in the queries to the Measurements API. This allows queries to be processes that contain series which are not present.

I opted to always keep this enabled as opposed to adding another flag to enable/disable it.


RangerRick:
LGTM, I'll test when you resolve confilcts with Dustin's merge.

RangerRick:
WFM

PR #90, opened 2019-02-27
HELM-131: fix template parsing with nested parens

RangerRick:
JIRA: https://issues.opennms.org/browse/HELM-131

Details here: https://opennms.discourse.group/t/helm-3-0-breaks-template-variables/220

This PR cleanly handles nested parentheses even when multiple template functions are discovered.  It does *not* handle nested functions.  I believe I can make that work, but spent too much time messing with it so I'll leave it for a future enhancement.

This PR also includes the fix from https://github.com/OpenNMS/opennms-helm/tree/grafana/pr/385 so I could test on both Grafana 5 and 6.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ awsconfig
 /tmp
 vendor/phantomjs/phantomjs
 vendor/opennms*.js*
+vendor/parenthesis
 vendor/crypto-js
 
 docs/AWS_S3_BUCKET
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -33,6 +33,10 @@ module.exports = function(grunt) {
         src: 'node_modules/crypto-js/md5.js',
         dest: 'vendor/crypto-js/md5.js'
       },
+      parenthesis_to_vendor: {
+        src: 'node_modules/parenthesis/index.js',
+        dest: 'vendor/parenthesis/index.js'
+      },
       vendor_to_dist: {
         cwd: 'vendor',
         expand: true,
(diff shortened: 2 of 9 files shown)


j-white:
Testing on Grafana 6.0.0:

`nodeFilter(nodeLabel like 'ellbat%')` calls:
http://localhost:3333/api/datasources/proxy/1/rest/nodes?filterRule=nodeLabel+like+%27ellbat%25%27&limit=0

which returns a 200

`nodeFilter((nodeLabel like 'ellbat%'))` calls:
http://localhost:3333/api/datasources/proxy/1/rest/nodes?filterRule=(nodeLabel+like+%27ellbat%25%27&limit=0

j-white:
(Scratch that last comment I wasn't running with these commits -_-)

j-white:
The alarm histogram panel still doesn't seem to be working on 6 for me. I don't get the error when it's loaded anymore, but the panel doesn't show any data.

RangerRick:
Hrm, OK, I only spot-checked that it loaded without throwing an error, I did most of my testing in 5...  I'll go back and take a look.

RangerRick:
OK, I force-pushed a change with that commit replaced with [something more robust](https://github.com/OpenNMS/opennms-helm/commit/11addedd6fb1ca697e37f6d2510c959869bdec1b).  I must have just gotten lucky with my first run because it definitely wasn't working for me this time around.

j-white:
Move commit related to Grafana 6 support to the `jira/HELM-132` branch.

PR #93, opened 2019-03-26
HELM-133: inventory datasource

RangerRick:
This PR adds an inventory datasource that queries nodes in the same tabular style that the fault datasource queries alarms.

It also includes a refactor of the underlying query style the `fault-ds` uses to better match Grafana convention.


j-white:
We already have the problem of having too many datasources, so I'm reluctant to add another one, especially when it closely resembles the existing fault-ds.

Short of exposing the queries necessary for the new filter panel, is there any use case for this datasource? Looking at the node details in a table panel doesn't need particularly useful.

If we do need it, would it be possible to refactor the existing fault-ds with this functionality instead?
In the query editor, we could make the `all alarms` field a dropdown that would allow you to switch between query inventory and alarms.

RangerRick:
The use case is not wanting to use the PostgreSQL datasource.  Inventory isn't a "fault" so it seems weird to me to stuff it into the fault-ds. I would also expect as an end-user of Helm, you may want to use the Inventory datasource for *other* things like having a searchable table of nodes in Grafana.

RangerRick:
Also it's only complicated because there's a lot of duplicate code that I want to clean up in another pass, but there really isn't *that* much to the logic.

j-white:
We can rename the fault-ds to something more generic if necessary.

Having a searchable table of nodes doesn't seem compelling - there's no state, nothing tied to the time range, and no place to get further details.

j-white:
Circling back here, I mostly want to make sure that we have a strategy for dealing with datasources going forward. I don't think it's feasibly to continue adding new datasources for every new type of data we want to expose to panels, particularly for cases where the settings are the same, and the query editor is very similar.

We may not be able to solve this immediately, but I wanted to point that out before we move forward with this.

RangerRick:
Alright, I _think_ I have this all fixed up and working.  I'd love some eyes on it to make sure everything looks good.

I deliberately split out the commit renaming `fault-ds` to `entity-ds` so that you can (hopefully) look at the subsequent commits and get a better idea of what's been done without all the noise from the renames.  It still has a _lot_ of changes because there's quite a bit of reworking needed to make this DS two-headed.  On the bright side, adding other entities should be easier going forward.

RangerRick:
Also note, I did document how to use `wizzy` to update your datasources without having to re-connect them to the new datasource, since this is a breaking change.

src/plugin.json
```diff
@@ -19,17 +19,18 @@
...
     ],
-    "version": "3.0.2-SNAPSHOT",
+    "version": "4.0.0-SNAPSHOT",
     "updated": "2019-03-27"
   },
   "includes": [
     {"type": "panel", "name": "Alarm Table"},
     {"type": "panel", "name": "Alarm Histogram"},
+    {"type": "panel", "name": "Filter Panel"},
     {"type": "datasource", "name": "OpenNMS Faults"},
```
j-white: Faults vs Entity

I'm guessing this is why ` OpenNMS Faults` still shows up in the list for `Add data source`.

PR #94, opened 2019-03-26
HELM-135: `isAcknowledged` enhancements

RangerRick:
* add `Is Acknowledged` to the data returned by the fault-ds
* add a new `Boolean` column type to the Alarm Table panel which
  can display Y/N/true/false/etc. as a checkbox or not

Is Acknowledged Column


RangerRick:
@pbrane has a separate enhancement for changing the default alarm table panels that includes a number of other tweaks.

PR #99, opened 2019-04-11
HZN-1492 Tag "root cause" alarm when providing feedback

smith-opennms:
This PR consists of the HELM UI work to allow a user to select ONE of the related alarms as the 'root cause' of the Situation.
This information is passed via opennms-js/ReST to the OpenNMS server and then persisted in ElasticSearch.
Additionally, a user can provide a list of 'Tags' that will be attached to each situation/alarm feedback record.

This PR also offers a different style on the feedback buttons in conjunction with the new RootCause button.

That said, there is probably lots of room for improvement, so all suggestions are welcome.

JIRA: https://issues.opennms.org/browse/HZN-1492

Currently, the UI looks like this:

image

### DIFF
--- a/src/panels/alarm-table/alarm_details.js
+++ b/src/panels/alarm-table/alarm_details.js
@@ -67,7 +67,7 @@ export class AlarmDetailsCtrl {
     }
 
     // Feedback Tags
-    $scope.existingFeedbackTags = new Set([]);
+    $scope.feedbackTags = new Set([]);
 
     // If this is a Situation, collect any correlation feedback previously submitted
     if ($scope.alarm.relatedAlarms && $scope.alarm.relatedAlarms.length > 0) {
@@ -202,9 +202,9 @@ export class AlarmDetailsCtrl {
 
   submitEditedFeedback(form) {
     for (let feedback of this.$scope.situationFeedback) {
+      feedback.tags = this.$scope.feedbackTags;
       if (form) {
         feedback.reason = form.reason;
-        feedback.tags = this.$scope.feedbackTags;
       }
     }
     this.submitFeedback(this.$scope.situationFeedback);
@@ -228,14 +228,14 @@ export class AlarmDetailsCtrl {
 
   updateFeedback(feedback) {
     for (let fb of feedback) {
-      for (let ifb of this.$scope.situationFeedback) {
-        if (fb.alarmKey === ifb.alarmKey)
-          ifb = fb;
-      }
+      const index = this.$scope.situationFeedback.findIndex(ifb => ifb.alarmKey === fb.alarmKey);
+      this.$scope.situationFeedback[index].rootCause = fb.rootCause;
+      this.$scope.situationFeedback[index].tags = fb.tags;
       for (let tag of fb.tags) {
-        this.$scope.existingFeedbackTags.add(tag);
+        this.$scope.feedbackTags.add(tag);
       }
     }
+    $('tags-input').tagsinput('refresh');
   }
 
   editSituationFeedback() {
--- a/src/panels/alarm-table/alarm_details.js
+++ b/src/panels/alarm-table/alarm_details.js
@@ -58,8 +58,9 @@ export class AlarmDetailsCtrl {
 
     // Situation Feedback
     $scope.situationFeebackEnabled = false;
+    $scope.feedbackReason = '';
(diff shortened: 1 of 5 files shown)


RangerRick:
My only concern is that with long log messages, that UI can already get pretty weird... the new columns take up a lot of space.  That said, I'm not sure how to make them smaller without making it a pull-down picker or something, which makes the "Tally" bit weird.  Also "tally" sounds awkward somehow, but I haven't thought of anything that works better yet.  Maybe "Totals"? or something?

smith-opennms:
On logline length, what about making the modal wider overall. should be fine except for those OCEs with 480x640 CRTs
The tally was put in during the last rev, when icons were used to indicate good and bad correlations (it wasn't obvious or intuitive). Perhaps it's no longer required.

RangerRick:
Wider modal makes sense to me... it's not like what's under it is terribly accessible.  Might as well make it as wide as we can.

j-white:
Gave this a whirl, here's what I found:

1. Minor visual "issues"

image

2. On a fresh install, open the dialog, click around related/not related, click cancel, modal doesn't dismiss. Console shows:
```
TypeError: Cannot read property 'length' of undefined
    at e.value (module.js:1231)
    at e.value (module.js:1231)
    at fn (eval at compile (angular.js:15551), <anonymous>:4:182)
    at i (angular.js:27341)
    at l.$eval (angular.js:18423)
    at l.$apply (angular.js:18523)
    at HTMLButtonElement.<anonymous> (angular.js:27346)
    at HTMLButtonElement.dispatch (jquery.js:5183)
    at HTMLButtonElement.g.handle (jquery.js:4991)
```

3. Tags and comment do not get re-populated after submitting and refreshing the page
4. Feedback button disappears if there's feedback, and a related alarm goes away. Fails with:

```

Situation Feedback not supported error:  TypeError: Cannot read property 'timestamp' of undefined
    at a (module.js:1231)
    at e.value (module.js:1231)
    at eval (module.js:1231)
    at angular.js:17051
    at angular.js:17095
    at l.$digest (angular.js:18232)
    at angular.js:18462
    at h (angular.js:6362)
    at angular.js:6642
```

j-white:
Some suggestions:

1. Add `border-radius: 0px;` to the `feedback-button` class.
2. My guess is that this is coming from the `resetCounters` call
3. Use a list and not a set for the `feedbackTags` - see https://github.com/bootstrap-tagsinput/bootstrap-tagsinput/blob/master/src/bootstrap-tagsinput-angular.js#L25
4. Validate that `this.$scope.situationFeedback.findIndex` returns an index > 0

src/panels/alarm-table/alarm_details.js
```diff
@@ -270,14 +270,16 @@ export class AlarmDetailsCtrl {
...
       const index = this.$scope.situationFeedback.findIndex(ifb => ifb.alarmKey === fb.alarmKey);
-      if (this.$scope.situationFeedback[index].timestamp < fb.timestamp) {
-        this.$scope.situationFeedback[index].rootCause = fb.rootCause;
-        this.$scope.situationFeedback[index].tags = fb.tags;
-        this.$scope.situationFeedback[index].feedbackType = fb.feedbackType;
-        this.$scope.situationFeedback[index].timestamp = fb.timestamp;
-        for (let tag of fb.tags) {
-          this.$scope.feedbackTags.add(tag);
-          this.$scope.tagsInput.tagsinput('add', tag);
+      if (index >= 0 && index <= this.$scope.situationFeedback.length) {
```
j-white: We don't need to check for `index <= this.$scope.situationFeedback.length`. 

I'd replace this the following check instead of indenting the whole block:
```
if (index < 0) {
        // We have feedback for an alarm that no longer exists in the situation, ignore it.
        continue;
      }
```

smith-opennms:
There are still 2 outstanding issues:

1) There is a weird behaviour when 
* the Tags are prepopulated, 
* and the user goes to resubmit feedback
* and deletes one or more tags 
* and then focus leaves the tags input
the tags return. if they are deleted again, they remain deleted.

2) Longer log messages wrap and cause the artifacts in the first image - where one can see the severity color behind (under) the button color. This might be mostly remedied with a wider modal area.

smith-opennms:
Here's how the changes are reflected in the UI
image

PR #100, opened 2019-05-16
HELM-147: wrap log/description/operinst

RangerRick:
This patch also does a couple of other small layout cleanups while I was in this code.

* make the modal dynamically re-size itself vertically (previously, it would set the max height at popup time, but then if the window was resized, it wouldn't adjust)
* make the "severity" row in the details view colorized
* fix the JSON tab to not waste extra space and do nicer scroll bar boundaries


RangerRick:
those attributes should have gotten removed from the top of the details page, I think you got the merge backwards... I'll check it tomorrow

PR #104, opened 2019-06-03
HELM-148: Control resolution when using the flow datasource

mattixtech:
- Allow user to override the step to group results by a given interval in string format (10s, 5m, 1h etc...)
- Added cardinality to functions to prevent those that are to be applied only once from being presented in the menu after already being applied
- Added mutual exclusions to functions to prevent conflicting functions from being applied (such as topN and withApplication)

This screenshot shows the new transform function to apply the interval `withGroupByInterval` and the function being excluded from the menu due to cardinality of single.
Screen Shot 2019-06-03 at 3 55 54 PM

* Temporarily based on top of the other drift2.0 work, will re-target when that is in

Jira: https://issues.opennms.org/browse/HELM-148

### DIFF
--- a/src/datasources/flow-ds/datasource.js
+++ b/src/datasources/flow-ds/datasource.js
@@ -44,7 +44,7 @@ export class FlowDatasource {
     let asTableSummary = FlowDatasource.isFunctionPresent(target, 'asTableSummary');
 
     // If a group by interval has been set we will use that to determine the step value, otherwise we will use the step
-    // value from grafan's automatically calculated maxDataPoints (based on pixel width)
+    // value from Grafana's automatically calculated maxDataPoints (based on pixel width)
     let groupByInterval = this.getFunctionParameterOrDefault(target, 'withGroupByInterval', 0, null);
     let step;
     if (groupByInterval) {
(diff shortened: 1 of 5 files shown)


src/datasources/flow-ds/datasource.js
```diff
@@ -359,7 +370,11 @@ export class FlowDatasource {
       return def;
     }
     // Return the parameter value, and perform any required template variable substitutions
-    return this.templateSrv.replace(func.parameters[idx]);
+    if (Gfuncs.getFuncDef(name).params[0].type === 'int') {
```
j-white:
```suggestion
    if (Gfuncs.getFuncDef(name).params[idx].type === 'int') {
```

mattixtech:
@j-white The current behaviour of the table summary options is that it does not apply any other transform functions. So I've made it mutually exclusive of those functions in the latest commit. Would any of these functions be desirable to have working for the table summary?

j-white:
We should get `toBits` to work the the summary too.

mattixtech:
> We should get `toBits` to work the the summary too.
Fixed in latest commit.

PR #105, opened 2019-06-13
HELM-145: Histogram panel for flows

mattixtech:
This PR adds a new histogram panel for visualizing flow data (by hosts, applications, conversations).

The only thing I'm not happy with is the legend. Using the flot legend plugin but I'm currently hardcoding the dimensions right now and styling it with a combination of stylesheets and javascript. Ideally I'd like to switch to the same type of legend that Grafana uses for the graph panel. However it is functional right now so that can be addressed later. I created https://issues.opennms.org/browse/HELM-156 to address the legend.

See screenshots below for the configuration supported.

Right now it is named 'Helm Histogram' but could stand to have a better name...

I've manually tested all the configurations against my local Opennms instance (which only has ingress data, hence the lack of egress in screenshot below).

Jira: https://issues.opennms.org/browse/HELM-145

Screen Shot 2019-06-13 at 11 15 42 AM

Screen Shot 2019-06-13 at 11 15 17 AM


j-white:
Nice work.

For the name, maybe something like  "Flow Histogram" would be sufficient.

A few visual nits:
1. Can we remove the white border around the legend color?
2. Can we add some padding between the elements in the legend? The end of the label from the previous touch the border of the next one.

package.json
```diff
@@ -55,6 +55,7 @@
     "babel-preset-es2015": "^6.24.1",
     "crypto-js": "^3.1.9-1",
     "flot": "https://github.com/flot/flot#v0.8.3",
+    "flot-axislabels": "https://github.com/markrcote/flot-axislabels#release-2.0.1",
```
RangerRick: FYI, we have a version of this with a proper `package.json` that makes node/yarn happier: https://github.com/j-white/flot-axislabels#master

mattixtech: Thanks, will use that one.

mattixtech:
Feedback addressed in latest commit. Here is a screenshot:

Screen Shot 2019-06-19 at 11 15 12 AM

mattixtech:
> Nice work.
> 
> For the name, maybe something like "Flow Histogram" would be sufficient.
> 
> A few visual nits:
> 
> 1. Can we remove the white border around the legend color?
> 2. Can we add some padding between the elements in the legend? The end of the label from the previous touch the border of the next one.

Done.

PR #107, opened 2019-06-18
HELM-137: add a filter panel

RangerRick:
This currently only works properly in conjunction with a panel
that is using the fault-ds.  It special-cases looking for the
filter panel and, if found, adds its criteria to its own.

This is a squash of previous commits made during HELM-137 dev:

* ULU-44: show the datasource in the editor
* ULU-45: fix support for using the same column multiple times
* ULU-51: auto-wildcard text fields
* HELM-142: fix missing filter panel text values on reload


mattixtech:
I don't really understand what this feature does, can you provide a screenshot or expand on the description please?

In what circumstances would this new panel be used?

Thanks

j-white:
Is the expectation that we can set restrictions on both alarms and nodes in the filter panel and have these filter a single "alarm table" panel with an alarm query? Or can we only use alarm filters for alarms?

Asking this since adding a filter for a "Node: Label" on "Node" entity, causes the alarm table to fail with: `GET http://localhost:3000/api/datasources/proxy/15/api/v2/alarms?_s=(lastEventTime%3Dge%3D2019-06-19T14%253A57%253A36.530%252B0000;lastEventTime%3Dle%3D2019-06-19T20%253A57%253A36.530%252B0000);(node.node.label%3D%3Dthresholding-test-node)&limit=0 500 (Internal Server Error)`

RangerRick:
Oop, good catch.  It should work, I think you found 2 different bugs, one in mapping, one in not-putting-prefixes-on.  :D

RangerRick:
Not been able to reproduce @j-white's issue, might have come from having an old config.  Otherwise I _think_ this is ready for merge.  I fixed the last bug I ran into with the picker.

j-white:
Tried out the latest build with Grafana 6.2.2, getting this error when clicking `+`:
image

src/panels/filter-panel/ctrl.js
```diff
@@ -0,0 +1,172 @@
...
+
+        this.events.on('init-edit-mode', this.onInitEditMode.bind(this));
+        this.events.on('data-received', this.onDataReceived.bind(this));
+        this.events.on('data-error', this.onDataError.bind(this));
+        this.events.on('data-snapshot-load', this.onDataReceived.bind(this));
+        this.events.on('render', this.onRender.bind(this));
+    }
+
+    link($scope, elem, attrs, ctrl) {
+        // this.elem = elem.find('.histogram-chart');
```
mattixtech:
```suggestion
```

src/panels/filter-panel/value_select_dropdown.js
```diff
@@ -0,0 +1,314 @@
+import angular from 'angular';
+import _ from 'lodash';
+
+import coreModule from 'app/core/core_module';
+
+export class OnmsValueSelectDropdownCtrl {
+  /*
+  dropdownVisible: any;
```
mattixtech: Is this for typescript?

RangerRick: Yes and no, it's because I ported the grafana internal `value_select_dropdown.ts` to pure javascript because I don't have access to the objects directly.  I left everything as in-place as possible and just commented some stuff out.

j-white:
Still failing with the same problem for me:
image

RangerRick:
huh, weird... is there more to that stack trace?

RangerRick:
@j-white lemme know when you've had a chance to update minion-dev with the latest build and I'll dig around and see if it's better.  Still WFM on my system.

RangerRick:
FYI, all my pending `jira/JS-*` stuff is in the `ranger/wip` branch I temporarily pointed this helm at, so this *should* be testable shortly.

j-white:
Working better now. Some initial comments:

1. Can we add some padding between the first "filter" and the left side of the panel?

image

2. When adding a new filter column the list of available filters only seems to show automatically when adding a column for the *first* time. For subsequent adds, I need to start typing for it to show up.

image

image

3. Would it be feasible to make the filter names configurable?

RangerRick:
1. done
2. I'm not convinced this isn't a bug in Grafana itself, I've seen this behavior too and not been able to narrow down why it's happening... I can take another look, but I think it's worth opening a new bug and not blocking merge for it though.
3. I think so... I'll have to see if it honors passing custom `value:` stuff through.  The datasource variable is a tricky beast.

RangerRick:
I've created issues for 2 & 3: [HELM-164](https://issues.opennms.org/browse/HELM-165), [HELM-165](https://issues.opennms.org/browse/HELM-165)

PR #108, opened 2019-06-19
HELM-159: add CSV and Excel export

RangerRick:
It should not have been this easy.  :D

JIRA: https://issues.opennms.org/browse/HELM-159


j-white:
Should this be "selection context aware"? Or do we just dump the whole table everytime?

RangerRick:
This just dumps the whole table.  Not sure it's likely they're going to want to dump only selected things, especially since I don't think selections cross page boundaries.

PR #110, opened 2019-06-20
Use the provided version of Flot instead of pulling in our own.

j-white:
This fixes a bug where the Pie Chart panel can't be used on the same dashboard as the Flow Histogram panel.

The histogram panel appears to be working properly after these changes.


j-white:
@mattixtech can you give this a try and confirm that the "Flow Histogram" panel continues to work as expected?

PR #112, opened 2019-07-10
HELM-162: Add legend controls to the "flow histogram" panel

mattixtech:
This PR updates the flow planel legend to use a legend adapted from Grafana's piechart plugin (https://github.com/grafana/piechart-panel). This legend allows for colors to be customized and individual series hidden from rendering.

The only thing I wasn't able to adapt cleanly when updating our legend is determining height (in the case of legend under chart) and width (in the case of legend beside chart) automatically. Instead I am using reasonable defaults and allow the values to be manually specified.

I've tested this against latest helm with latest opennms.

Please find screen capture of the new behaviour here: https://recordit.co/kwRwqfQceG

Jira: https://issues.opennms.org/browse/HELM-162


pbrane:
Wow!  This is some nice work on the panel, @mattixtech.

mattixtech:
> Wow! This is some nice work on the panel, @mattixtech.

Thanks! 😄

j-white:
Looks good overall.

Can we reduce the spacing the colors and the labels in the legend? (Seems large compared to other panels.)

image

mattixtech:
> Looks good overall.
> 
> Can we reduce the spacing the colors and the labels in the legend? (Seems large compared to other panels.)
> 
> image

I'll take a look at that. I'm using the same CSS as the piechart plugin so the spacing is the same. When it is in table format the spacing is reduced. If you toggled "show legend values" in your example you would end up with the same spacing as well.

mattixtech:
Padding has been reduced
Screen Shot 2019-07-15 at 5 51 22 PM

PR #186, opened 2019-09-03
HELM-175: Use hasOwnProperty instead of _.has to improve performance

j-white:
JIRA: https://issues.opennms.org/browse/HELM-175

With these changes processing 5000 alarms with 100 event parameters each
goes from taking 67631ms to 799ms (85x faster).


pbrane:
From > a minute to < a second, that’s pretty good! :)

RangerRick:
LGTM.

In general we should probably be trying harder to avoid using lodash (which is huge and hard to tree-shake) in favor of using newer builtin JavaScript features.  Babel already contains performant polyfills if we need to be more backwards-compatible.

PR #247, opened 2019-11-08
convert yarn build to npm

RangerRick:
nothing special, just doing the convert here to avoid an extra dependency


docs/installation/requirements.adoc
```diff
@@ -3,10 +3,9 @@
 == Runtime Requirements
 
 * http://docs.grafana.org/installation[Grafana]
-The required Grafana-Server version is *4.4.1* up to *< 6*.
+The required Grafana-Server version is *6.4.0* or later.
```
mvrueden: not sure why this was changed here

RangerRick: Just happened to notice it while I was updating the docs. I can change it in the other branch instead.

PR #254, opened 2020-01-07
Qauseenmz/helm bug fixes

QauseenMZ:
JIRA ISSUE: HELM-201

- Panel refresh after all clear actions are successful
- Executing clear action batch synchronously

### DIFF
--- a/src/datasources/entity-ds/datasource.js
+++ b/src/datasources/entity-ds/datasource.js
@@ -362,7 +362,7 @@ export class OpenNMSEntityDatasource {
 
     clearAlarm(alarmId) {
         const clearAlarmDefer = this.q.defer();
-        this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then((successObj) => {
+        this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then(() => {
             clearAlarmDefer.resolve({status: true})
         }, (error) => {
             console.log("DataSource Clear Alarm Error: ", error)
--- a/src/datasources/entity-ds/datasource.js
+++ b/src/datasources/entity-ds/datasource.js
@@ -361,12 +361,7 @@ export class OpenNMSEntityDatasource {
     }
 
     clearAlarm(alarmId) {
-        return this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then(() => {
-            return {status: true};
-        }, (error) => {
-            console.log("DataSource Clear Alarm Error: ", error)
-            return {status: false, error: error}
-        });
+        return this.q.when(this.opennmsClient.doClear(alarmId, this.user))
     }
 
     escalateAlarm(alarmId) {
--- a/src/datasources/entity-ds/datasource.js
+++ b/src/datasources/entity-ds/datasource.js
@@ -361,7 +361,14 @@ export class OpenNMSEntityDatasource {
     }
 
     clearAlarm(alarmId) {
-        return this.opennmsClient.doClear(alarmId, this.user);
+        const clearAlarmDefer = this.q.defer();
+        this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then((successObj) => {
+            clearAlarmDefer.resolve({status: true})
+        }, (error) => {
+            console.log("DataSource Clear Alarm Error: ", error)
+            clearAlarmDefer.reject({status: false, error: error})
+        });
+        return clearAlarmDefer.promise;
     }
 
     escalateAlarm(alarmId) {
--- a/src/datasources/entity-ds/datasource.js
+++ b/src/datasources/entity-ds/datasource.js
@@ -353,19 +353,19 @@ export class OpenNMSEntityDatasource {
     }
 
     acknowledgeAlarm(alarmId) {
-        return this.q.when(this.opennmsClient.doAck(alarmId, this.user));
+        return this.opennmsClient.doAck(alarmId, this.user);
     }
 
     unacknowledgeAlarm(alarmId) {
-        return this.q.when(this.opennmsClient.doUnack(alarmId, this.user));
+        return this.opennmsClient.doUnack(alarmId, this.user);
     }
 
     clearAlarm(alarmId) {
-        return this.q.when(this.opennmsClient.doClear(alarmId, this.user))
+        return this.opennmsClient.doClear(alarmId, this.user);
     }
 
     escalateAlarm(alarmId) {
-        return this.q.when(this.opennmsClient.doEscalate(alarmId, this.user));
+        return this.opennmsClient.doEscalate(alarmId, this.user);
     }
 
     createTicketForAlarm(alarmId) {
--- a/src/datasources/entity-ds/datasource.js
+++ b/src/datasources/entity-ds/datasource.js
@@ -361,14 +361,12 @@ export class OpenNMSEntityDatasource {
     }
 
     clearAlarm(alarmId) {
-        const clearAlarmDefer = this.q.defer();
-        this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then(() => {
-            clearAlarmDefer.resolve({status: true})
+        return this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then(() => {
+            return {status: true};
         }, (error) => {
             console.log("DataSource Clear Alarm Error: ", error)
-            clearAlarmDefer.reject({status: false, error: error})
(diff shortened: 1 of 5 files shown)


src/datasources/entity-ds/datasource.js
```diff
@@ -361,7 +361,14 @@ export class OpenNMSEntityDatasource {
...
     clearAlarm(alarmId) {
-        return this.opennmsClient.doClear(alarmId, this.user);
+        const clearAlarmDefer = this.q.defer();
+        this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then(() => {
+            clearAlarmDefer.resolve({status: true})
+        }, (error) => {
+            console.log("DataSource Clear Alarm Error: ", error)
+            clearAlarmDefer.reject({status: false, error: error})
+        });
+        return clearAlarmDefer.promise;
```
RangerRick: `doClear` is already a non-angularjs Promise; I think you should be able to just do this.q.when() to wrap it in an angular Promise without having to do the extra `defer` stuff.  (`q.when` will auto-convert any "then-able" promise to a Q promise.)  If you still want to return a "status" object you should be able to do this:

```javascript
return this.q.when(this.opennmsClient.doClear(alarmId, this.user)).then(() => {
  return { status: true };
}).catch((err) => {
  console.log("DataSource Clear Alarm Error:", error);
  return { status: false, error: error };
});
```

src/panels/alarm-table/action_mgr.js
```diff
@@ -119,10 +127,17 @@ export class ActionMgr {
...
-        // Apply the action to each row in the selection
-        _.each(rows, row => action(row));
+        // Apply the action(s) to each row in the selection, one after another on success
+        async.eachLimit(rows, 1, (row, callback) => action(row, callback), (error) => {
+          if(error){
+            //Log Error and terminate action(s) for other row(s)
+            console.log('Error in clearing: ', error)
+          }
+          else {
+            this.ctrl.refreshDashboard();
```
RangerRick: Could it be a problem that the dashboard is getting refreshed multiple times in quick succession?  Is it enough to refresh when the `eachLimit` calls are finished?

src/panels/alarm-table/module.js
```diff
@@ -687,23 +687,34 @@ class AlarmTableCtrl extends MetricsPanelCtrl {
...
   performAlarmActionOnDatasource(source, action, alarmId) {
     let self = this;
-    this.datasourceSrv.get(source).then(ds => {
-      if (ds.type && ds.type.indexOf("entity-datasource") < 0) {
-        throw {message: 'Only OpenNMS datasources are supported'};
-      } else {
-        if (!ds[action]) {
-          throw {message: 'Action ' + action + ' not implemented by datasource ' + ds.name + " of type " + ds.type};
+    
+    return new Promise((resolve, reject) => {
```
RangerRick: Is there a reason you're making a new `Promise` when `this.datasourceSrv.get()` already starts a promise chain?  You could just return it...

src/panels/alarm-table/action_mgr.js
QauseenMZ: Because for one action, all the panel data is being fetched after each iteration so just to make less frequent server request we discussed that refreshing in the end should be fine.

src/datasources/entity-ds/datasource.js
QauseenMZ: Thats right! I got it now. Thanks for mentioning! Updated the pull request for this in commit: https://github.com/OpenNMS/opennms-helm/pull/254/commits/46fb262bc1269eb2659846ff59f9aafaa8d482ab

src/panels/alarm-table/action_mgr.js
RangerRick: Right, I just meant that if you are processing a series of actions then it's going to refresh over and over in quick succession.  Is that refreshed UI useful to the user, or should you wait to refresh until all actions have completed?

src/panels/alarm-table/module.js
QauseenMZ: Yes `performAlarmActionOnDatasource` wasn't returning a promise necessarily (in other conditions), making .then() undefined for `performAlarmActionOnDatasource`.. so wrapping it in a new promise helped in creating the chain..

src/panels/alarm-table/action_mgr.js
j-white: We want all of the actions to be executed in parallel and once they are all complete, the dashboard should be refreshed.

src/panels/alarm-table/module.js
RangerRick: Ah, ok, thanks.

src/panels/alarm-table/action_mgr.js
RangerRick: AFAICT `async.eachLimit` with a limit of `1` means they won't run in parallel, doesn't it?  Or are the actions already running as promises so they happen at the same time anyway?

QauseenMZ: That's right.. its going one after another. Updated the pull request by making the actions request in parallel and not terminate if error with any of them.
Dashboard refreshing after the last action is executed.
Thanks!

src/datasources/entity-ds/datasource.js
```diff
@@ -361,7 +361,12 @@ export class OpenNMSEntityDatasource {
     }
 
     clearAlarm(alarmId) {
```
j-white: The escalate, acknowledge and un-acknowledge actions (available via the right click context menu) should also be updated to function similarly.

QauseenMZ: Other actions working in similar pattern because `performAlarmActionOnDatasource` is common for them. `clearAlarm` doesn't need to handle promise necessarily.. updated the PR in this commit https://github.com/OpenNMS/opennms-helm/pull/254/commits/622a8e48a7860dfa4305cdf0810952919bb74e7a.
Thanks!! :)

src/datasources/entity-ds/datasource.js
```diff
@@ -353,19 +353,19 @@ export class OpenNMSEntityDatasource {
...
     }
 
     unacknowledgeAlarm(alarmId) {
-        return this.opennmsClient.doUnack(alarmId, this.user);
+        return this.q.when(this.opennmsClient.doUnack(alarmId, this.user));
     }
 
     clearAlarm(alarmId) {
-        return this.opennmsClient.doClear(alarmId, this.user);
+        return this.q.when(this.opennmsClient.doClear(alarmId, this.user))
```
j-white:
```suggestion
        return this.q.when(this.opennmsClient.doClear(alarmId, this.user));
```

src/lib/client_delegate.js
```diff
@@ -184,7 +184,7 @@ export class ClientDelegate {
     }
 
     doClear(alarmId, user) {
-        return this.getAlarmDao()
```
j-white: Necessary change?

src/panels/alarm-table/action_mgr.js
```diff
@@ -31,28 +32,57 @@ export class ActionMgr {
     // We should only ack alarms that are not already acked
     let acknowledgeableRows = _.filter(this.rows, row => row.alarm.ackTime === void 0);
     this.addOptionToContextMenu('General', 'Acknowledge', acknowledgeableRows,
-        (row) => self.ctrl.acknowledgeAlarm(row.source, row.alarmId));
+        (row, callback) => {
+          self.ctrl.acknowledgeAlarm(row.source, row.alarmId).then(() => {
+            callback(null)
```
j-white: Add `;` to complete statements for consistency.

(Here and other lines as well.)

src/datasources/entity-ds/datasource.js
j-white: Can't we use the existing promised returned by `doClear` instead of creating a new one?

QauseenMZ: Yes, we should use the same.. this is extra now. Updated the PR in https://github.com/OpenNMS/opennms-helm/pull/254/commits/2f63d6b3d3b28c3a09a77158f3843b0d246f6138
Thanks! :)

src/panels/alarm-table/action_mgr.js
QauseenMZ: Updated: https://github.com/OpenNMS/opennms-helm/pull/254/commits/5eb67e89dac946f0c4def6b219c2e0f7824745cb

PR #255, opened 2020-01-08
HELM-202: Sort meta-data along with rows

j-white:
JIRA: https://issues.opennms.org/browse/HELM-202


src/spec/alarm_table_model_spec.js
```diff
@@ -0,0 +1,63 @@
+import {TableModel} from '../panels/alarm-table/table_model';
+
+describe('TableModel', function() {
+
+  it('should work', function () {
```
RangerRick: LOL

j-white: Forget I left that there... I guess I'll keep it for lolz :)

PR #257, opened 2020-01-09
Qauseenmz/helm-64

QauseenMZ:
HELM-64: Node search allows only to select the first 25 nodes	
Fix: Pagination Added

### DIFF
--- a/src/datasources/perf-ds/modal_ctrl.js
+++ b/src/datasources/perf-ds/modal_ctrl.js
@@ -3,27 +3,41 @@ class ModalCtrl {
   constructor($scope) {
     this.$scope = $scope;
     this.query = "";
+    this.offset = 0;
     this.selectedRow = null;
-
+    this.currentPage = 0;
+    this.pageSize = 25;
     this.searchForRows();
   }
 
   searchForRows() {
     var self = this;
     this.searching = true;
-    this.$scope.search(this.query)
+    
+    this.$scope.search(this.query, this.offset)
       .then(function (results) {
         // Reset the selected row
         self.selectedRow = null;
         // Add the results to the scope
         self.rows = results.rows;
+        self.allRows = results.rows;
         self.count = results.count;
         self.totalCount = results.totalCount;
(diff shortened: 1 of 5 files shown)


src/datasources/perf-ds/modal_ctrl.js
```diff
@@ -36,6 +50,24 @@ class ModalCtrl {
     }
   }
 
+  nextPage(){
+    if((this.currentPage + 1) == this.$scope.ctrl.numberOfPages){
+      return;
+    }
+    this.offset += 25;
```
j-white: We should re-use the page size constant instead of using `25` here.

src/datasources/perf-ds/modal_ctrl.js
```diff
@@ -36,6 +50,24 @@ class ModalCtrl {
...
+    this.currentPage++;
+    this.searchForRows();
+  }
+
+  prevPage(){
+    if(this.currentPage == 0){
+      return;
+    }
+    this.offset -= 25;
```
j-white: Similarly here.

src/datasources/perf-ds/modal_ctrl.js
```diff
@@ -36,6 +50,24 @@ class ModalCtrl {
     }
   }
 
+  nextPage(){
+    if((this.currentPage + 1) == this.$scope.ctrl.numberOfPages){
+      return;
+    }
+    this.offset += 25;
+    this.currentPage++;
+    this.searchForRows();
```
j-white: In the case where the given search function does not support the offset and always returns the full results, this will cause all of the data to be fetched again - which we don't want.

QauseenMZ: Thanks for mentioning! Client-side pagination added: https://github.com/OpenNMS/opennms-helm/pull/257/commits/9dbc78d3ba0977c483cc565571e965576ced4416

j-white:
Looks good.

I think we'll have to change the "Showing X of Y" label:
image

For example, when we're on the 2nd page, we're still only showing 25 resources, but the label seems to indicate that 50 results are shown.

PR #270, opened 2020-02-13
HELM-214/HELM-222: Publish build artifacts with CircleCI to Cloudsmith

indigo423:
# Pull Request Check Sheet

* [x] Have you read and followed our [Contribution Guidelines](https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md)?
* [x] Have you [made an issue in the OpenNMS issue tracker](https://issues.opennms.org)?<br>If so, you should:
  1. update the title of this PR to be of the format: `${JIRA-ISSUE-NUMBER}: subject of pull request`
  2. update the JIRA link at the bottom of this comment to refer to the real issue number
  3. prefix your commit messages with the issue number, if possible
* [x] Have you made a comment in that issue which points back to this PR?
* [x] Have you updated the JIRA link at the bottom of this comment to link to your issue?
* [ ] If this is a new feature, is there documentation?
* [ ] If this is a new feature or substantial change, are there tests that cover it?

Created two build jobs to build DEB and RPM packages. Added the Cloudsmith orb to publish it to our [repository](https://cloudsmith.io/~opennms/repos/common/packages/). The Cloudsmith ORB requires the CLOUDSMITH_API_KEY environment variable to be set on the project.

## Reviewer hint

We create RPM and DEB packages in the `dist/packages` directory. In the `make-deb` and `make-rpm`, I've just set to store everything in the dist/packages dir see:

```yaml
      - store_artifacts:
          path: ./dist/packages
```
There is no wildcard support in the store_artifacts directive and I didn't want to mess around the makerpm and makedeb scripts.

# External References

* JIRA (Issue Tracker): https://issues.opennms.org/browse/HELM-214
* JIRA (Issue Tracker): https://issues.opennms.org/browse/HELM-222

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,9 @@ executors:
     docker:
       - image: opennms/netlify-cli:2.8.3-b1
 
+orbs:
+  cloudsmith: cloudsmith/cloudsmith@1.0.3
+
 commands:
   docker-registry-login:
     description: "Connect to Docker Registry"
@@ -273,6 +276,32 @@ jobs:
           name: Deploy docs to Netlify
           command: netlify deploy --prod -d dist/docs/helm -s ${NETLIFY_SITE_ID}
 
+  publish-rpm:
+    executor: cloudsmith/default
+    steps:
+      - attach_workspace:
+          at: ~/
+      - cloudsmith/ensure-api-key
+      - cloudsmith/install-cli
+      - cloudsmith/publish:
+          cloudsmith-repository: opennms/common
+          package-format: rpm
+          package-distribution: any-distro/any-version
+          package-path: dist/packages/*.rpm
+
+  publish-deb:
+    executor: cloudsmith/default
+    steps:
+      - attach_workspace:
+          at: ~/
+      - cloudsmith/ensure-api-key
+      - cloudsmith/install-cli
+      - cloudsmith/publish:
+          cloudsmith-repository: opennms/common
+          package-format: deb
+          package-distribution: any-distro/any-version
+          package-path: dist/packages/*.deb
+
 workflows:
   version: 2
   build-workflow:
@@ -340,5 +369,25 @@ workflows:
             branches:
               only: 
                 - master
+            tags:
+              ignore: /^v.*/
+      - publish-deb:
+          requires:
+            - make-deb
+          filters:
+            branches:
+              only: 
+                - master
+                - jira/HELM-214
+            tags:
+              ignore: /^v.*/
+      - publish-rpm:
+          requires:
+            - make-rpm
+          filters:
+            branches:
+              only: 
+                - master
+                - jira/HELM-214
             tags:
               ignore: /^v.*/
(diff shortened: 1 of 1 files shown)


RangerRick:
Should it be a hard requirement that we support signing before we merge this for real?

indigo423:
> Should it be a hard requirement that we support signing before we merge this for real?

absolutely not. We can add this here as well.

indigo423:
@RangerRick I've made an issue for the signing in Helm ([HELM-222](https://issues.opennms.org/browse/HELM-222)). I've seen you created an issue for the signing orb, I've linked the issue as a dependency.

indigo423:
@RangerRick no objections to be the test 🐰 for your orb :)

.circleci/config.yml
```diff
@@ -238,6 +276,32 @@ jobs:
...
+          package-format: rpm
+          package-distribution: any-distro/any-version
+          package-path: dist/packages/*.rpm
+
+  publish-deb:
+    executor: cloudsmith/default
+    steps:
+      - attach_workspace:
+          at: ~/
+      - cloudsmith/ensure-api-key
```
j-white: We could consider using CircleCI contexts to populate the env. variables so we don't need to repeat them in every project.

.circleci/config.yml
```diff
@@ -270,6 +334,18 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+      - make-rpm:
```
j-white: What's the intended logic with the RPM/Deb building/publishing?

Build packages in all branches, but only push for tags starting with `v` and `master`?

indigo423: @j-white @RangerRick We do releases right now manually we haven't really talked about how we want to have the release logic. I would suggest following the approach from Alec where we publish release artifacts on git version tags prefixed with `v`, develop and release branches:

```yaml
filters:
  tags:
    only: /^v.*/
      branches:
        only:
          - develop
          - /release-.*/
```
It gives us a little bit more flexibility. We could do also something more restrictive and just publish release artifacts on git version tags:

```yaml
filters:
  tags:
    only: /^v.*/
      branches:
        ignore: /.*/
```
I'm open for one way or another. We can decide this here in the comments and fix the PR. I'll extract the release workflow from this conversation and we can it to the docs.

indigo423:
@j-white @RangerRick I was able to get the signing with your orb to work and everything worked as expected. Do we want to decide in HELM-214 what kind of release workflow we want to have or should we go ahead and move it to a dedicated issue?

j-white:
> @j-white @RangerRick I was able to get the signing with your orb to work and everything worked as expected. Do we want to decide in HELM-214 what kind of release workflow we want to have or should we go ahead and move it to a dedicated issue?

I think we should avoid pushing RPMs for every branch initially - maybe just develop for now and then we can refine it?

RangerRick:
yeah, I would only do develop for now... we don't really have support for "branch" packages in cloudsmith anyway

indigo423:
> yeah, I would only do develop for now... we don't really have support for "branch" packages in cloudsmith anyway

@j-white @RangerRick filter for publish-rpm and publish-deb is just set to develop. Should be good then so we can move forward and merge it. Any objections?

PR #273, opened 2020-04-15
HELM-232: fix Grafana 6.7 support

RangerRick:
This PR updates to OpenNMS.js 2.0.2, which fixes an issue where sometimes the `config` property on the response object is not set when Grafana passes it back into our code.


grafana.version
```diff
@@ -1 +1 @@
-6.6.1
+6.7.1
```
indigo423, line 1: Meanwhile, it is 6.7.2, should we risk it :)

RangerRick: 6.7.2 is what I tested on, honestly, I'll bump it after I merge.

PR #277, opened 2020-07-20
HELM-243: fix attribute node search modal in Grafana 6.7

RangerRick:
This PR fixes the attribute node search popup to render properly and to react when searches happen. The fix is to make sure everything is happening in Angular promises and the chains complete, and also for the search input to be selected when the modal opens.

@agalue I think this fixes your issue, if you don't mind testing it out.

https://issues.opennms.org/browse/HELM-243


RangerRick:
Also, good news!  I tried it in Grafana 7.1 and it still works there.

RangerRick:
:D Yeah, still some work to do for 7, but I was glad it still worked there.

PR #279, opened 2020-08-05
Jeffg/devjam2020/ship more dashboards

jeffgdotorg:
# Pull Request Check Sheet

* [x] Have you read and followed our [Contribution Guidelines](https://github.com/OpenNMS/opennms/blob/develop/CONTRIBUTING.md)?
* [x] Have you [made an issue in the OpenNMS issue tracker](https://issues.opennms.org)?<br>If so, you should:
* [x] Have you made a comment in that issue which points back to this PR?
* [x] Have you updated the JIRA link at the bottom of this comment to link to your issue?
* [N/A] If this is a new feature, is there documentation?
* [N/A] If this is a new feature or substantial change, are there tests that cover it?

# Pull Request Process

One or more reviewers should be assigned to each PR.

If you know that a particular person is subject matter expert in the area your PR affects, feel free to assign one or more reviewers when you create this PR, otherwise reviewers will be assigned for you.

Once the reviewer(s) accept the PR and the branch passes continuous integration in Bamboo, the PR is eligible for merge.

At that time, if you have commit access (are an OpenNMS Group employee or a member of the Order of the Green Polo) you are welcome to merge the PR.
Otherwise, a reviewer can merge it for you.

Thanks for taking time to contribute!

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-249
* Continuous Integration: [Bamboo](https://bamboo.opennms.org/), [CircleCI](https://circleci.com/gh/OpenNMS/opennms-js)

### DIFF
--- a/src/dashboards/current_alarms.json
+++ b/src/dashboards/current_alarms.json
@@ -701,7 +701,7 @@
     ]
   },
   "timezone": "",
-  "title": "Current Alarms",
+  "title": "Current Alarms | OpenNMS Helm",
   "uid": "GOCJnbnGz",
   "version": 5
 }
--- a/src/dashboards/current_alarms.json
+++ b/src/dashboards/current_alarms.json
@@ -60,7 +60,7 @@
           "id": "7VM5VtsBi5T5Qsu2TybX",
           "text": "Category: Name",
           "label": "Node Category",
-          "datasource": "opennms-entity",
+          "datasource": "${DS_OPENNMS-ENTITY}",
           "resource": "category.name",
           "inputType": "multi",
           "entityType": {
@@ -69,7 +69,7 @@
             "queryFunction": "alarms"
           },
           "selected": {
-            "datasource": "opennms-entity",
+            "datasource": "${DS_OPENNMS-ENTITY}",
             "entityType": {
               "id": "alarm",
               "label": "Alarms",
@@ -87,7 +87,7 @@
           "id": "oZF0CsfaTbARufEUIemG",
           "text": "Node: Label",
           "label": "Node Label",
-          "datasource": "opennms-entity",
+          "datasource": "${DS_OPENNMS-ENTITY}",
           "resource": "node.label",
           "inputType": "text",
           "entityType": {
@@ -96,7 +96,7 @@
             "queryFunction": "alarms"
           },
           "selected": {
-            "datasource": "opennms-entity",
+            "datasource": "${DS_OPENNMS-ENTITY}",
             "entityType": {
               "id": "alarm",
               "label": "Alarms",
--- a/src/dashboards/current_alarms.json
+++ b/src/dashboards/current_alarms.json
@@ -1,0 +1,707 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_OPENNMS-ENTITY",
+      "label": "opennms-entity",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "opennms-helm-entity-datasource",
+      "pluginName": "OpenNMS Entities"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "6.3.6"
+    },
+    {
+      "type": "panel",
+      "id": "opennms-helm-alarm-table-panel",
+      "name": "Alarm Table",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "opennms-helm-entity-datasource",
+      "name": "OpenNMS Entities",
+      "version": "1.0.0"
+    },
+    {
+      "type": "panel",
+      "id": "opennms-helm-filter-panel",
+      "name": "Filter Panel",
+      "version": ""
+    }
+  ],
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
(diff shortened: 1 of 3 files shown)


mfuhrmann:
It's a good idea to have some default boards right out of the box :+1:

I would like to share here my experience about dashboards since I spend a lot of time with my colleagues in that topic. Not sure if these setups work for the broad mass.

### Related to the alarm board

I would suggest to add a datasource chooser. Often you have an OpenNMS test system or multiple instances in general. With the datasource chooser you can easily choose between your OpenNMS instances instead of creating a new board that points to an other system.
Looks like this in Grafana:
image

Having all alarms in one box is not useful in the field in my opinion. Especially when you have a lot of alarms, the normal alarm box is not scaling well due to this per page view mechanism. And exactly in this situations OpenNMS is a very good tool to keep an overview (but that is also due to the 24h availability views and the separated alarm views on the starting page in my opinion).
In my setup for example I filtered out the BSM alarms from the normal alarm box to and added an own alarm box only for BSM alarms. As operator you will have a better overview.
Depending on which features the customer is using (and of course that's the crux), it might also made sense to separate out the situation alarms. We do it in a similar way on the starting page. This is very handy to have.
Also I'm not a fan of having resolved events in the alarms view. So I usually filter alarms `> NORMAL`. Either I have active alarms or not. Also the paging issue mentioned above can hit you here very hard. 

I really like the Node Label field, but I would rename it to "Node Filter", because it was not clear for me on the first view that it is a "live filter" for the list.
It would be cool if this filters would work in an URL. Often you want to point a colleague to an alarm in Grafana and I didn't find a way to do that. Often colleagues have Grafana access but not OpenNMS access.

I also have the IP address as a column in the alarm box. In my case it makes sense. 

### Related to the node metrics board

What I've noticed in your board is that there is nothing about the memory. Also the CPU load is missing which is really important.
System uptime is also often nice to have to see when a reboot happened. The mentioned datasource chooser above and I would suggest to add the all options for the filesystem, disk and interface chooser.

I'm having trouble with the interface filters on `(eth|en|tun)`. Are those the only ones that are officially "well known"? The drop down list I get is very straightforward.
In my setups it would not work. We have `bond`, `vlan` or `vmbr` which are not displayed because of this filter. But of course those are more or less manually created and not provided by the Linux installation.
Maybe we can ignore some well known bad asses here to get the list shorter but not restricted on the 3 above?
I use a interface ignore policy in my OpenNMS that ignores: `docker|tap|veth|fwbr|fwln|fwpr`

I personally only use only [this one](https://grafana.com/grafana/dashboards/5148) for SNMP Linux nodes and dropped all my own boards, since it contains nearly everything SNMP provides and works out of the box. And in my opinion we should maintain this board officially.

I would like to raise the question here : Why do you want to have a board containing node metrics? - My answer: To be able to analyze strange behaviors. Everything else is done with alarms and already known. But in this case I need as many metrics as possible to be able to draw conclusions.

---

Sorry for being so picky here :-)

indigo423:
there is not much I can add here, very nice write up Mr. @mfuhrmann :)

mfuhrmann:
Also I want to mention, that the [board](https://grafana.com/grafana/dashboards/5148) I suggested above, also misses some things. For example memory usage in % or it does not use the cpuCount metric.

@jeffgdotorg Would be happy if you give my improved version of the one above a try.
[LinuxNodePerformanceBoard.zip](https://github.com/OpenNMS/opennms-helm/files/5101509/LinuxNodePerformanceBoard.zip)

The only thing that is not working in my setup in this board are the pie charts when I use RRD strategy. I get an `internal server error`. When using Newts it works. But it might be an issue with my RRD setting. You should double check that.

mmahacek:
Bump on this PR.  I'm working on https://github.com/OpenNMS/opennms-helm/pull/300 to refresh the existing Flow dashboard.  Having some other built-ins would be helpful for users new to Helm.

mmahacek:
What about an "informational" dashboard for new users to get started with Helm the first time?

image

jeffgdotorg:
@mmahacek there was some conversation (not captured in this PR) that led me to abandon the whole effort to ship more dashboards. I forget the details but it boiled down to a problem with the way Grafana manages installed dashboard JSON when Grafana itself gets updated, and left me convinced that shipping additional substantial boards in Helm would burden us with a giant mess. Perhaps Grafana has changed how they do this and it's time to reevaluate, but I haven't kept up.

An "info" board in the box, though, is something I can +1.

mmahacek:
That makes sense. I have been working on updating the default Flow dashboard as it was last updated for Grafana 5.  I will say it did seem to import properly into 7.5, though when I export it back out to make updates, there definitely are a good number of fields that are updated on the panels (though nothing breaking).  The dashboard JSON in the plugin doesn't look too different from dashboards uploaded to the galleries.  There may be a point where the extra repos will have the same issue with older dashboard formats.
And I can totally agree that everyone's environment is different, so it's tricky to determine which kinds of dashboards should be included out of the box. Anything added in the plugin would add additional review for verifying on future releases and wouldn't be easy for users to tweak without risk of overwriting them on Helm update.

Probably can just close out this issue.

mmahacek:
https://issues.opennms.org/browse/HELM-281 - Create a Welcome dashboard for new Helm users

PR #282, opened 2020-09-30
HELM-247: fix alarm table and filter on Grafana 7

RangerRick:
This is incredibly gross, but I got it all working on Grafana 6 as well as Grafana 7 (I tested with 6.7.3, 7.1.0 and 7.2.0)

* port over variable-management code from Grafana 6.7
* disable ad-hoc queries in Grafana 7 (I _think_ we don't actually mark any columns as filterable anyway, so this should be a no-op, in theory)
* cleanups to the filter panel layout to fit Grafana 7's right-side tray format
* debounce input to text fields in the filter panel

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-247

### DIFF
--- a/src/lib/filter_column.js
+++ b/src/lib/filter_column.js
@@ -1,3 +1,6 @@
+import { TextBoxVariable } from './text_box_variable';
+import { QueryVariable } from './query_variable';
+
 const randomString = () => {
   const chars = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXTZabcdefghiklmnopqrstuvwxyz'.split('');
   const length = 20;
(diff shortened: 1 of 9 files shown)


agalue (approved):
I compiled the branch and tested it against Grafana 7.2.0, and it works as expected. Here is what I tried:

* The Alarm Table
* The Alarm Histogram
* Looking at the details of a given alarm
* Follow the link to OpenNMS from the details modal
* Acknowledge alarms
* Add a sticky note
* Add a filter panel with a text field, and make sure the other widgets are updated when applying a filter

RangerRick:
Quick, gonna merge it before anything else can go wrong.

PR #283, opened 2020-10-22
HELM-258: feat(tos): add support to filter by ToS

fooker:
Adds

* TypeScript support to Helm
* `withDscp()` and `withEcn()` filter
* multi-selection template variables for DSCPs and ECNs
* two rows (for DSCP and ECN) that show series and summaries
* `dscpOnExporterNodeAndInterface()` and `ecnOnExporterAndInterface()` queries

This pull request requires an update to OpenNMS-JS (see https://github.com/OpenNMS/opennms-js/pull/205).

This also includes updates to the bundled dashboard (See HELM-253, HELM-255, HELM-256, HELM-257).

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-258


RangerRick:
Looks OK.  One question: if I understand the code correctly, it looks like you're basically always asking for tos but using a default of `0` if nothing is specified.  Is this backwards-compatible with older OpenNMS releases?  Do flow queries against Horizon 26 or Meridian 2020 still work with these changes in place?

fooker:
> Looks OK. One question: if I understand the code correctly, it looks like you're basically always asking for tos but using a default of `0` if nothing is specified. Is this backwards-compatible with older OpenNMS releases? Do flow queries against Horizon 26 or Meridian 2020 still work with these changes in place?

Haven't tested it against an old OpenNMS version yet. My plan was to do so after the other changes related to the ToS topic are done (maybe we should merge this into a feature branch instead?)

My understanding that `getFunctionParameterOrDefault` will return `undefined` if the filter is not used. This will be passed to an URL query parameter. And therefore the parameter will be omitted if the filter is not used. Even if it is used, older versions will simply ignore the parameter.

fooker:
@swachter I've seen some `Template variables could not be initialized: For input string: " "` errors while giving this a try. Looks like this happens, if there is no known interface for the exporter yet (which can happen during initialization / reprovisioning.

fooker:
As I can't really do reviews on my own PR, I have to do it inline...

@swachter I don't like the restriction of the template queries to the selected time-range. For multiple reasons:
* The selection is inconsistent when the time-selection changes. What happens to the selection if the user drills down into a section where a value is no longer present? Do I have to recheck my filter every time I zoom out?
* This breaks all existing templates. As the parameters are not optional, users have to change all templates (or re-import and loosing their customization) after updating.
* This was not part of the task/issue this PR should implement.

In addition, I also don't understand some other seemingly unrelated changes? What does the changes to the build scripts do and why are they required? If this is just for convenience, I would like to see them in a separate commit with some detailed explanation.

Last but not least a minor thing: Why has `intervalToMs` been moved around?

swachter:
> As I can't really do reviews on my own PR, I have to do it inline...
> 
> @swachter I don't like the restriction of the template queries to the selected time-range. For multiple reasons:
> 
>     * The selection is inconsistent when the time-selection changes. What happens to the selection if the user drills down into a section where a value is no longer present? Do I have to recheck my filter every time I zoom out?

This is an interesting one! I relied on the comment written in `opennms-js` for `FlowDao.getExporter` and `FlowDao.getExporters`. There it is stated that:

```
     * @param start - the start of the timespan to query (defaults to 4 hours ago)
     * @param end - the end of the timespan to query (defaults to now)
```

If that comment would be true and Grafana's time range would not considered then only exporters / interfaces from the last 4 hours would be available for selection.

However, I checked the implementations of `FlowRestServiceImpl.getExporter()` and `FlowRestServiceImpl.getExporters()`. Apparently they do not respect `start` and `end` parameters given in requests. They use the time range [now - `OnmsSnmpInterface.MAX_FLOW_AGE`, now] that defaults to the last 168 days and can be configured by the system property `org.opennms.features.telemetry.maxFlowAgeSeconds`.

In summary, the changed implementation has no influence on the selected exporter / interfaces.

I think the available exporter / interfaces (and dscps) should be based on the selected time range in future. Grafana offers two choices when variables are refreshed: "On Dashboard Load" or "On Time Range Changed". Both choices are valid options. I would default to "On Time Range Changed". I think the question if template variables should depend on the selected time range is independent of when they should be refreshed.

Assuming template variables are refreshed "On Time Range Changed": I expect that Grafana preserves the value of a template variable when you zoom out because the available exporter / interfaces in the the larger time range also include the current values. I am not sure of the behavior when you zoom in, though. In that case the current values may no more included in the available values. I did not find documentation about this situation and had no data at hand to check the behavior. I would expect that Grafana just keeps the current values.


> 
>     * This breaks all existing templates. As the parameters are not optional, users have to change all templates (or re-import and loosing their customization) after updating.

I had a discussion with @RangerRick  on this. He had no objections in a breaking change there. He said: "I wouldn't consider helm code as an "API" in any way".

I think I should enhance the regular expression logic such that start / end are optional. Would this cover your concerns?


> 
>     * This was not part of the task/issue this PR should implement.

Agreed. It seemed to be an obvious and quick win (if the 4h default would take effect if the time range is not considered).


> 
> 
> In addition, I also don't understand some other seemingly unrelated changes? What does the changes to the build scripts do and why are they required? If this is just for convenience, I would like to see them in a separate commit with some detailed explanation.

The changes to the build scripts add TypeScript support to the build. I needed 4 commits until everything was in place. Unfortunately, these commits also contained some "feature" related changes. It would take me some substantial amount of time to 
reorder and cherry pick these things. I have to improve my Git skills to accomplish this. Please confirm if you really need this.


> 
> Last but not least a minor thing: Why has `intervalToMs` been moved around?

Ben had moved that function into `utils.js`. Unfortunately, this broke the development build. Grafana could not load some module code at runtime via `SystemJS`. After quite some time trying to resolve the issue Ben suggested just to move `intervalToMs` to the datasource again.

RangerRick:
To be clear, I don't consider helm's internals to be an API, but if template variables that are stored in dashboards and such change, I would consider that to be a thing to avoid breaking if possible.  It _is_ important that current helm still works against old OpenNMS versions, and that people's existing custom Helm dashboards in Grafana don't break.

fooker:
> Agreed. It seemed to be an obvious and quick win (if the 4h default would take effect if the time range is not considered).

Even if things seem like a "quick win" I consider it bad behavior to sneak them into other PRs. Just create a separate one and have a tracking issue for it.

> In summary, the changed implementation has no influence on the selected exporter / interfaces.

Then there is no need to add these parameters.

> The changes to the build scripts add TypeScript support to the build.

This is also fare beyond the scope of this issue and makes it super hard to review thing. Starting with the fact that I will not decide to add another language to a project without talking to some more people. This is way to much effort for a little helper function.

swachter:
> > Agreed. It seemed to be an obvious and quick win (if the 4h default would take effect if the time range is not considered).
> 
> Even if things seem like a "quick win" I consider it bad behavior to sneak them into other PRs. Just create a separate one and have a tracking issue for it.
> 
> > In summary, the changed implementation has no influence on the selected exporter / interfaces.
> 
> Then there is no need to add these parameters.

Ok. I remove all of that.

> 
> > The changes to the build scripts add TypeScript support to the build.
> 
> This is also fare beyond the scope of this issue and makes it super hard to review thing. Starting with the fact that I will not decide to add another language to a project without talking to some more people. This is way to much effort for a little helper function.

Before I introduced TypeScript into the Helm project I started a discussion in the Engineering channel on December 10th. Matt and Ben were supportive and there was no objection.

RangerRick:
Definitely support using TypeScript, but it should definitely be in a separate PR or at least separate standalone commit.

I've broken out the TS stuff into a separate commit (and changed a few things to make it happy with a project that doesn't _contain_ any TS yet) and merged to `develop`, so you should be able to rebase this to just be pure code changes related to the feature now.

swachter:
rebase on `develop`

src/lib/utils2.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * Converts an input string into a string array of selected values.
+ *
+ * The input string may be the value of a multi-selection template variable. In this case the input string is
+ * delimited by curly braces that enclose a comma separated list of values.
+ *
+ * The dropUnresolved parameter determines if an empty array is returned if the input starts with '$'.
+ * The dropAll parameter determines if an empty array is returned if the input is "all".
+ */
+export function processSelectionVariable(input: string, dropUnresolved: boolean, dropAll: boolean): string[] {
```
RangerRick, line 10: Can the code in `utils2.ts` be merged into `utils` now? I presume it was a separate file just so you could write it in typescript.

swachter: Yes, they can be merged.

RangerRick (approved):
Honestly thought I had already approved this. LGTM, thanks!

PR #295, opened 2021-04-30
HELM-267: fix initialization of Helm components

RangerRick:
This PR converts the entire Helm project to:

1. use `@grafana/toolkit` for the build system
2. get rid of all "compatibility layer" work, and instead rely on `@grafana/data` and `@grafana/runtime` wherever possible
3. convert everything to typescript

This fixes a myriad of issues and provides new opportunities like signing plugins easily.

There are a HUGE number of changes.  Frankly, I'm not sure it's even useful to do a "code review" -- it's more useful to do actual testing of all components with live data and make sure things work.

If you _do_ want to review the code, I recommend checking it out and doing:

```
git diff -b -w '-M25%' origin/develop
```

...to force it to be super forgiving when detecting renames.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-267


swachter:
In the flow-deep-dive dashboard query segments can no more be edited. In earlier version one could click on segements and edit parameters. That is no longer possible. For example if a `topK` combiner is added, its default parameter of `10` can not be changed.

j-white:
Some observations while testing:

1) When adding a Perf DS, if the user/pass is not set, it shows this when saving/testing:
image

I would expect a red x / fail, for a bad request?

2) Filter panel

A lot of weirdness here: 
image

3) Alarm table overflows:

image

image

j-white:
Aside from the issues reported above, everything else seems to work, tested:
* Alarm table panel
   * w/ custom actions
* Flow histogram panel
* Perf DS
* Flow DS
* Entity DS

RangerRick:
Thanks, think I've fixed everything.

1. For the filter panel issue, the properties on the datasource are slightly different than they used to be, I changed the way the datasource list is generated and it should be better now.
2. For the layout issues, it was a problem with not always initializing CSS, I've fixed that by making a global "set up CSS if it isn't" method and called it from all modules.
3. The green "Bad Request" must have always been broken in that way. The `perf-ds` is the only thing not using OpenNMS.js/`lib/client_delegate` for making requests, and does all its own handling.  I've fixed it to properly pass through errors as promise rejections.

swachter (approved):
Well done! I tested the flow-deep-dive dashboard.

PR #315, opened 2021-11-29
HELM-301: get rid of angular.IQService and angular.IPromise

swachter:
Issue: https://issues.opennms.org/browse/HELM-301

Grafana no more requires the use of special Angular promises; standard promises can be used.

I did some tests with the flow data source. However, the other functionalities of helm have still to be tested.


RangerRick (approved):
@swachter this needs some conflict resolution/updating but otherwise I'm cool with doing it

PR #319, opened 2021-12-14
HELM-293: support string properties in performance datasource

swachter:
Issue: https://issues.opennms.org/browse/HELM-293

This PR adds support for querying string properties to the performance datasource.

@Bonrob2 I also added some documentation; please review.

Open question: I added some help text into `query.options.html` but did not find where this would show up when using the performance datasource. Is the help functiionality broken?

### DIFF
--- a/docs/modules/datasources/pages/performance_datasource.adoc
+++ b/docs/modules/datasources/pages/performance_datasource.adoc
@@ -41,12 +41,19 @@ Use Grafana's transforms like "Filter by name", "Organize Fields", or "Merge" to
 
 [options="header, autowidth"]
 |===
-| Name               | Description
-| Node               | Node ID of the node to query.
+| Name
+| Description
+
+| Node
+| Node ID of the node to query.
 Can also be a variable with a `nodeFilter()` query.
-| Resource           | Resource ID containing the requested attribute.
+
+| Resource
+| Resource ID containing the requested attribute.
 Can also be a variable with a `nodeResources()` query.
-| String Property    | Name of the attribute to query.
+
+| String Property
+| Name of the attribute to query.
 |===
 
 === Expression queries
--- a/docs/modules/datasources/pages/performance_datasource.adoc
+++ b/docs/modules/datasources/pages/performance_datasource.adoc
@@ -33,7 +33,11 @@ The query is used to identify which attribute (or series) is being queried.
 
 === String property queries
 
-Used to read the last value of string valued properties that OpenNMS collects and stores. The query is used to identify which string property is queried. A panel can contain multiple string property queries. The result can be visualized in a table. Grafana's transforms like "Filter by name", "Organize Fields", or "Merge" can be used to adjust the visualized table as needed.
+Used to read the last value of string-valued properties that OpenNMS collects and stores.
+The query identifies which string property is queried. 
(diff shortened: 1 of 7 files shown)


docs/modules/datasources/pages/performance_datasource.adoc
```diff
@@ -31,6 +31,20 @@ The query is used to identify which attribute (or series) is being queried.
                        Must be unique for all attributes and expression results in the query.
 |===
 
+=== String property queries
+
+Used to read the last value of string valued properties that OpenNMS collects and stores. The query is used to identify which string property is queried. A panel can contain multiple string property queries. The result can be visualized in a table. Grafana's transforms like "Filter by name", "Organize Fields", or "Merge" can be used to adjust the visualized table as needed.
```
Bonrob2, line 36:
```suggestion
Used to read the last value of string-valued properties that OpenNMS collects and stores.
The query identifies which string property is queried. 
A panel can contain multiple string property queries. 
The result can be visualized in a table. 
Use Grafana's transforms like "Filter by name", "Organize Fields", or "Merge" to adjust the visualized table as needed.
```
Each sentence needs to be on a separate line.

Bonrob2 (changes requested):
Looks good, Stefan. Minor changes.

RangerRick (approved):
this looks pretty good to me 👍

docs/modules/datasources/pages/performance_datasource.adoc
```diff
@@ -31,6 +31,20 @@ The query is used to identify which attribute (or series) is being queried.
...
+
+[options="header, autowidth"]
+|===
+| Name               | Description
+| Node               | Node ID of the node to query.
+Can also be a variable with a `nodeFilter()` query.
+| Resource           | Resource ID containing the requested attribute.
+Can also be a variable with a `nodeResources()` query.
+| String Property    | Name of the attribute to query.
+|===
```
mmahacek, line 46:
```suggestion
|===
| Name
| Description

| Node
| Node ID of the node to query.
Can also be a variable with a `nodeFilter()` query.

| Resource
| Resource ID containing the requested attribute.
Can also be a variable with a `nodeResources()` query.

| String Property
| Name of the attribute to query.
|===
```

RangerRick (approved):

Bonrob2 (dismissed):
LGTM!

mmahacek (changes requested):
Modified Signature error with deb on Ubuntu with Grafana 8.3.3

mmahacek:
And the ZIP from the latest build running on 8.3.3 in Docker has an issue where UI elements aren't displayed. 
image

RangerRick (changes requested):
please rebase this to `release-7.x`

swachter:
rebased on 7.x; please review again

mmahacek (approved):
This works for me on Grafana 8.3.3 on both Ubuntu and Docker.

RangerRick (dismissed):

swachter:
I am sorry, I have to make another small change. Simply throwing an error in case that string property queries are mixed with other kinds of queries does not yield to an error message in the view. The fix is to return a proper response with an error message.

RangerRick (approved):
ah yeah, good catch

PR #336, opened 2022-01-13
HELM-303: Use select resources endpoint when retrieving string properties and endpoint is available

swachter:
Issue: https://issues.opennms.org/browse/HELM-303


RangerRick (approved):
LGTM

PR #373, opened 2022-03-17
AR/jira/HELM-305: Helm's Flow datasource cannot filter "NaN" results, breaking Grafana reduce expressions

aramosvizcarra:
I believe the issue was a bug in the flows datasource for series(Nan values were actual string NaN, so Number.isNaN was retruning incorrect values). There is no need to create a Transformation for filter NaN unless is for an especific scenario where this grouping is needed.

Reduce Expression should be able to use Number type values from the flow series datasoiurces.

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-305

### DIFF
--- a/docs/modules/datasources/pages/flow_datasource.adoc
+++ b/docs/modules/datasources/pages/flow_datasource.adoc
@@ -25,6 +25,7 @@ The query editor supports the following functions:
 | Transform | `withGroupByInterval`   | Changes the resolution of the returned datapoints grouping by the given time interval (in the format of: 10s, 5m, 1h, etc.).
 | Transform | `withPrefix`            | Specifies a prefix for graph and summary labels.
 | Transform | `withSuffix`            | Specifies a suffix for graph and summary labels.
+| Transform | `nanToZero`             | Convert all NaN values to zero (0) so they can be taken in consideration by reduce expressions.
 |===
 
 
--- a/src/datasources/flow-ds/datasource.ts
+++ b/src/datasources/flow-ds/datasource.ts
@@ -389,6 +389,7 @@ export class FlowDatasource {
     let combineIngressEgress = FlowDatasource.isFunctionPresent(query, 'combineIngressEgress');
     let onlyIngress = FlowDatasource.isFunctionPresent(query, 'onlyIngress');
     let onlyEgress = FlowDatasource.isFunctionPresent(query, 'onlyEgress');
+    let nanToZero = FlowDatasource.isFunctionPresent(query, 'nanToZero');
 
     let start = flowSeries.start.valueOf();
     let end = flowSeries.end.valueOf();
@@ -424,8 +425,8 @@ export class FlowDatasource {
                       .filter(({column}) => column.label === col.label)
                       // get the values of those columns ...
                       .map(({colIdx}) => {
-                        const v = Number(values[colIdx][timestampIdx])
-                        return isNaN(v) ? 0 : v
+                        const v = values[colIdx][timestampIdx]
+                       return isNaN(Number(v)) && nanToZero ? 0 : v 
                       })
                       // ... and sum them up
                       .reduce((previous, current) => previous + (current ? current : 0), 0)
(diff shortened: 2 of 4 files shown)


UberPinguin:
When Grafana applies a Reduce expression to a data set, if _any_ of the values in that set are null or NaN, the entire expression returns NaN.
https://grafana.com/docs/grafana/latest/panels/query-a-data-source/use-expressions-to-manipulate-data/about-expressions/#reduce

j-white:
Mapping NaN (no data) to 0 is a bit deceptive: I would expect no point on the graph for NaN values as opposed to a point with a value of 0. If this behavior is required for grouping operations and reduce transforms, then it could be made "opt-in" via a transform.

UberPinguin:
> Mapping NaN (no data) to 0 is a bit deceptive: I would expect no point on the graph for NaN values as opposed to a point with a value of 0. If this behavior is required for grouping operations and reduce transforms, then it could be made "opt-in" via a transform.

If there's no data for a conversation key in a time bucket, wouldn't that indicate that the traffic during that interval was 0?
In what case would it be otherwise? Or, alternatively, if the traffic for that conversation pair in that interval were 0, what would you expect the flow exporter to have sent?

aramosvizcarra:
So I did a bit of testing based on @j-white  comment because I was under a similar thought as @UberPinguin 
Leaving null instead of 0 doesn't break the reduce expressions, but return different results like avg, max and min values since it only considere values that are not null.

I think this would be the correct representation if we assume no data "NaN" means that they should be excluded from the reduce expressions. 

See images below

Using null instead of 0 for NaN

Max
Screen Shot 2022-03-18 at 12 22 03 PM

Min
Screen Shot 2022-03-18 at 12 34 07 PM

Null
Screen Shot 2022-03-18 at 12 22 24 PM

Using 0 instead of null
Max 
Screen Shot 2022-03-18 at 12 39 21 PM

No longer Min
Screen Shot 2022-03-18 at 12 26 30 PM

New Min (0 instead of NaN)
Screen Shot 2022-03-18 at 12 34 58 PM

Based on that I'd would say to use null instead of 0 and add a transformation to convert NaN to 0 instead of null.

UberPinguin:
> Final results With "NaN" -> Null values Screen Shot 2022-03-22 at 8 57 21 AM
> 
> Using nanToZero transform Screen Shot 2022-03-22 at 9 00 47 AM

Am I missing something, or do the two screenshots show different values? Are they for the same timeframe?

aramosvizcarra:
> > Final results With "NaN" -> Null values Screen Shot 2022-03-22 at 8 57 21 AM
> > Using nanToZero transform Screen Shot 2022-03-22 at 9 00 47 AM
> 
> Am I missing something, or do the two screenshots show different values? Are they for the same timeframe?

They are different.

The first one is the OOTB behavior which wouldn't consider NaN (null) values in the reductions.

The second one, use the nanToZero transform that convert NaN to 0 values so Min, Max and Avg (or any other reduction) might show different values since is taking all NaN as 0 values.
 
So the "nanToZero" tranform is optional and can be used whenever someone wants to consider NaN as 0.

j-white:
Looks good. Can also document the transform here: https://github.com/OpenNMS/opennms-helm/blame/develop/docs/modules/datasources/pages/flow_datasource.adoc#L27

RangerRick (approved):

Bonrob2:
@aramosvizcarra when there are documentation changes, could you please tag me as a reviewer? There is a minor typo in the docs as merged.

docs/modules/datasources/pages/flow_datasource.adoc
```diff
@@ -25,6 +25,7 @@ The query editor supports the following functions:
 | Transform | `withGroupByInterval`   | Changes the resolution of the returned datapoints grouping by the given time interval (in the format of: 10s, 5m, 1h, etc.).
 | Transform | `withPrefix`            | Specifies a prefix for graph and summary labels.
 | Transform | `withSuffix`            | Specifies a suffix for graph and summary labels.
+| Transform | `nanToZero`             | Convert all NaN values to zero (0) so they can be taken in consideration by reduce expressions.
```
Bonrob2, line 28: Convert all NaN values to zero (0) so they can be taken into consideration by reduce expressions.

Also, not clear on what a reduce expression is. Should it be "reduced" expressions?

Bonrob2 (commented):
This had been merged, but there are two issues in the docs.

docs/modules/datasources/pages/flow_datasource.adoc
Bonrob2: Thanks! I'm adding these changes to HELM-315, so we can update this as part of other Helm doc changes.

PR #387, opened 2022-04-04
HELM-309: Grafana Datasource expressions - Performance

aramosvizcarra:
Added functionlity to allow Glob expressions ('*' and '|' )in order to query multiple resources and attributes from the performance datasource using a single query. (See images below with result samples)

Screen Shot 2022-04-04 at 4 58 34 PM

Screen Shot 2022-04-04 at 5 01 52 PM

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-309

### DIFF
--- a/src/datasources/perf-ds/datasource.ts
+++ b/src/datasources/perf-ds/datasource.ts
@@ -1,11 +1,13 @@
-import {QueryType, STRING_PROPERTY_TYPE} from './constants';
-import {interpolate} from "./interpolate";
+import { QueryType, STRING_PROPERTY_TYPE } from './constants';
+import { interpolate } from "./interpolate";
 import _ from 'lodash';
-import {FunctionFormatter} from '../../lib/function_formatter';
-import {DataQuery, DataQueryRequest, DataQueryResponse, Field, FieldType} from "@grafana/data";
-import {DataQueryResponseData} from "@grafana/data/types/datasource";
-import {ClientDelegate} from '../../lib/client_delegate'
-import {Client, ServerMetadata} from 'opennms'
+import { FunctionFormatter } from '../../lib/function_formatter';
+import { DataQuery, DataQueryRequest, DataQueryResponse, Field, FieldType } from "@grafana/data";
+import { DataQueryResponseData } from "@grafana/data/types/datasource";
+import { ClientDelegate } from '../../lib/client_delegate'
+import { Client, ServerMetadata } from 'opennms'
+
+const lodashClonedeep = require('lodash.clonedeep');
 
 interface Query {
   start: number;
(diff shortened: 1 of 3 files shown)


RangerRick (approved):
This looks fine to me.  A couple of notes in general to make things easier in the future:

* you can use `async`/`await` in Helm and it will compile down to "regular" promises for you, which would simplify a bunch of the `.then()` chains you're doing
* in most cases, if you're using underscore (especially `_.each()` and such) you can replace that with modern javascript `.forEach()` without having to pull it in

That said, those things aren't showstoppers to merging, just an FYI for future code. It can make things a bit more readable when you get into complicated promise workflows like this.

src/datasources/perf-ds/datasource.ts
```diff
@@ -316,6 +338,196 @@ export class OpenNMSDatasource {
...
+    return props;
+  }
+
+  /**
+   * Retrievd the node database id value from the resource id
+   * @param resource : resource id as <nodeId>.<resourceId>
+   * @returns Node Id 
+   */
+  static getNodeId(resource): string {
+    const matches = resource.match(/node(Source)?\[(.*)?\]\..*/);
```
RangerRick, line 511: Like I mentioned in https://github.com/OpenNMS/opennms-helm/pull/395, you might want to make this not do a "greedy anything" match:
```suggestion
    const matches = resource.match(/node(Source)?\[([^\]*)?\]\..*/);
```

src/datasources/perf-ds/datasource.ts
```diff
@@ -316,6 +338,196 @@ export class OpenNMSDatasource {
...
+    else { return resource; }
+  }
+
+  /**
+   * Gets the resource Id part from the resourceId property in a source object.
+   * @param resource : resource id as node[Source]<id>.<resourceId>
+   * @returns Resource Id
+   */
+  static getResourceId(resource): string {
+    const matches = resource.match(/node(Source)?\[.*?\]\.(.*)/);
```
RangerRick, line 524: same non-greedy
```suggestion
    const matches = resource.match(/node(Source)?\[[^\]]*?\]\.(.*)/);
```

aramosvizcarra:
> This looks fine to me. A couple of notes in general to make things easier in the future:
> 
> * you can use `async`/`await` in Helm and it will compile down to "regular" promises for you, which would simplify a bunch of the `.then()` chains you're doing
> * in most cases, if you're using underscore (especially `_.each()` and such) you can replace that with modern javascript `.forEach()` without having to pull it in
> 
> That said, those things aren't showstoppers to merging, just an FYI for future code. It can make things a bit more readable when you get into complicated promise workflows like this.

Will definetely  do, thanks for the feedback!

RangerRick (approved):

PR #389, opened 2022-04-12
HELM-307: Grafana Location Dashboard - Performance

aramosvizcarra:
Added a template variable nodeLocation that should allow norrow down nodes by location. This is meant to used together with nodeFilter so the selectable nodes are filtered by location.

 
Screen Shot 2022-04-11 at 1 13 08 PM

Screen Shot 2022-04-11 at 1 16 21 PM

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-307

### DIFF
--- a/docs/modules/datasources/pages/performance_datasource.adoc
+++ b/docs/modules/datasources/pages/performance_datasource.adoc
@@ -88,9 +88,9 @@ The results will appear as a drop-down list at the top of the dashboard.
 You can then select from the list to change the data the panel displays.
 See xref:panel_configuration:dynamic-dashboard.adoc#pc-template-filters[Using Template Variables to Create Dynamic Dashboards] for an example.
 
-=== nodeLocation()
+=== locations()
 
-The `nodeLocation()` query returns a list of all available node locations. 
+The `locations()` query returns a list of all available node locations. 
 Use this variable with nodeFilter to reduce the number of nodes displayed by location.
 
 === nodeFilter(<filterQuery>)
--- a/docs/modules/datasources/pages/performance_datasource.adoc
+++ b/docs/modules/datasources/pages/performance_datasource.adoc
@@ -88,6 +88,11 @@ The results will appear as a drop-down list at the top of the dashboard.
 You can then select from the list to change the data the panel displays.
 See xref:panel_configuration:dynamic-dashboard.adoc#pc-template-filters[Using Template Variables to Create Dynamic Dashboards] for an example.
 
+=== nodeLocation()
+
+The `nodeLocation()` query will return a list of all available nodes locations. 
+This variable is meant to be used together with nodeFilter, in order to reduce the number of nodes displayed by location.
+
 === nodeFilter(<filterQuery>)
 
 The `nodeFilter()` query will return a list of nodes which match the given filter expression.
--- a/docs/modules/datasources/pages/performance_datasource.adoc
+++ b/docs/modules/datasources/pages/performance_datasource.adoc
@@ -40,6 +40,7 @@ The result can be visualized in a table.
 Use Grafana's transforms like "Filter by name", "Organize Fields", or "Merge" to adjust the visualized table as needed.
 
 [options="header, autowidth"]
+[cols="1,2"]
 |===
 | Name
 | Description
--- a/docs/modules/datasources/pages/performance_datasource.adoc
+++ b/docs/modules/datasources/pages/performance_datasource.adoc
@@ -91,7 +91,7 @@ See xref:panel_configuration:dynamic-dashboard.adoc#pc-template-filters[Using Te
 === nodeLocation()
 
 The `nodeLocation()` query returns a list of all available node locations. 
-This variable is meant to be used together with nodeFilter, in order to reduce the number of nodes displayed by location.
+Use this variable with nodeFilter to reduce the number of nodes displayed by location.
 
 === nodeFilter(<filterQuery>)
 
--- a/docs/modules/datasources/pages/performance_datasource.adoc
+++ b/docs/modules/datasources/pages/performance_datasource.adoc
@@ -90,7 +90,7 @@ See xref:panel_configuration:dynamic-dashboard.adoc#pc-template-filters[Using Te
 
 === nodeLocation()
 
-The `nodeLocation()` query will return a list of all available nodes locations. 
+The `nodeLocation()` query returns a list of all available node locations. 
 This variable is meant to be used together with nodeFilter, in order to reduce the number of nodes displayed by location.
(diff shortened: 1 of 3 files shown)


aramosvizcarra:
I requested some feedback from @Will regarding creating a default dashboard as well in which case I would need to know what else I should include.

docs/modules/datasources/pages/performance_datasource.adoc
```diff
@@ -88,6 +88,11 @@ The results will appear as a drop-down list at the top of the dashboard.
 You can then select from the list to change the data the panel displays.
 See xref:panel_configuration:dynamic-dashboard.adoc#pc-template-filters[Using Template Variables to Create Dynamic Dashboards] for an example.
 
+=== nodeLocation()
```
fooker, line 91: IMO, this should be plural. It does not give the location of a node but all relevant location**s**.

src/datasources/perf-ds/datasource.ts
```diff
@@ -341,6 +343,29 @@ export class OpenNMSDatasource {
     return Promise.resolve([]);
   }
 
+  metricFindNodesLocation(searchLimit) {
```
fooker, line 346: The pluralization here is even more weird. Should be `metricFindLocations`.

docs/modules/datasources/pages/performance_datasource.adoc
fooker: Thinking some more, this should be `locations` only. By wording, there is no "Node Location h but only a "Location" where Nodes (and other stuff) are located in.

docs/modules/datasources/pages/performance_datasource.adoc
```diff
@@ -88,6 +88,11 @@ The results will appear as a drop-down list at the top of the dashboard.
...
+
+The `nodeLocation()` query will return a list of all available nodes locations.
```
Bonrob2, line 93:
```suggestion
The `nodeLocation()` query returns a list of all available node locations. 
```

docs/modules/datasources/pages/performance_datasource.adoc
```diff
@@ -88,6 +88,11 @@ The results will appear as a drop-down list at the top of the dashboard.
 You can then select from the list to change the data the panel displays.
 See xref:panel_configuration:dynamic-dashboard.adoc#pc-template-filters[Using Template Variables to Create Dynamic Dashboards] for an example.
 
+=== nodeLocation()
+
+The `nodeLocation()` query will return a list of all available nodes locations. 
+This variable is meant to be used together with nodeFilter, in order to reduce the number of nodes displayed by location.
```
Bonrob2, line 94:
```suggestion
Use this variable with nodeFilter to reduce the number of nodes displayed by location.
```

src/datasources/perf-ds/datasource.ts
```diff
@@ -341,6 +343,29 @@ export class OpenNMSDatasource {
...
+      }
+      var results = [] as any[];
+      _.each(response.data.node, function (node) {
+        var nodeLocation = node.location.toString();
+        if(nodeLocation){
+        results.push({text: nodeLocation, value: nodeLocation, expandable: true});
+        }
+      });
+      return results;
+    });
```
fooker, line 366: This seems to do the wrong thing. There is a one-to-many relation between location and node. Therefore this would add the same location multiple times if multiple nodes are assigned to the same location.

src/datasources/perf-ds/datasource.ts
```diff
@@ -341,6 +343,29 @@ export class OpenNMSDatasource {
     return Promise.resolve([]);
   }
 
+  metricFindNodesLocation(searchLimit) {
+    return this.doOpenNMSRequest({
+      url: '/rest/nodes',
```
fooker, line 348: There is a `monitoringLocations` REST API which should give you the requested information without provoking a full table scan on the nodes table.

fooker (changes requested):

Bonrob2 (changes requested):
minor changes to the docs.

Bonrob2:
@aramosvizcarra, separate from the changes you made, I noticed that the table in the "String property queries" does not display properly. Is this something you could fix as part of this work? I think if we add a [cols="1,2"] under the existing table syntax at line 42, it should work.

src/datasources/perf-ds/datasource.ts
aramosvizcarra: I was goint to filter but grafana already does it. I can pass 200 'Default' locations and only one shows up.

Still it is confusing, so I'll change it.

src/datasources/perf-ds/datasource.ts
aramosvizcarra: I was wondering about that. Looked though the documentation and couldn't find one. I'll fix that. Thanks!

will:
> I requested some feedback from @will regarding creating a default dashboard as well in which case I would need to know what else I should include.

I don't know what this project is, but if you really want my feedback: first off, there aren't any units on the y axis. 100000 what? Also that is wayyy to many zeros to count, make it like 10K or 1M. I didn't count how many zeros.

Are those times in local time or UTC? people are going to be asking that every time.

Bonrob2 (commented):
Doc changes LGTM!

docs/modules/datasources/pages/performance_datasource.adoc
aramosvizcarra: Applied change

src/datasources/perf-ds/datasource.ts
aramosvizcarra: Applied change

src/datasources/perf-ds/datasource.ts
aramosvizcarra: switechd to rest monitoringLocations which seems to hae unique values, but added a check to ensure no duplicates are passed.

src/datasources/perf-ds/datasource.ts
aramosvizcarra: switched rest call to use monitoringLocations instead

fooker (dismissed):

aramosvizcarra:
@Bonrob2 I added the format you mentioned on your last comment. If you could take a look and approve it is is all good that would be great!. Thank you.

Bonrob2 (dismissed):
Perfect. Thank you for fixing that!

fooker (dismissed):

RangerRick (dismissed):
looks good!

aramosvizcarra:
@RangerRick could you take one more look at this PR.

PR #412, opened 2022-05-04
HELM-320: FlowDS's combineIngressEgress transform loses the first query response result

aramosvizcarra:
Possible fix for missing records when using combineIngressEgress transformation

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-320

### DIFF
--- a/src/datasources/flow-ds/datasource.ts
+++ b/src/datasources/flow-ds/datasource.ts
@@ -441,18 +441,16 @@ export class FlowDatasource {
     }
 
     if (combineIngressEgress) {
-      const ingressColumns =  flowSeries.columns.filter(column => column.ingress);
-      const egressColumns =  flowSeries.columns.filter(column => !column.ingress);
-      const columns = ingressColumns.length >= egressColumns.length ? ingressColumns : egressColumns;
-
-      return columns          
+      const uniqueColumns =  [...new Set(flowSeries.columns.map(col => col.label))];
+      
+      return uniqueColumns          
           .map(col => {
 
             const datapoints = timestampsInRange
                 .map(({timestamp, timestampIdx}) => {
                   const sum = columnsWithIndex
                       // determine the indexes of those columns that have the same label as the current column
-                      .filter(({column}) => column.label === col.label)
+                      .filter(({column}) => column.label === col)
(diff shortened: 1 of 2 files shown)


src/datasources/flow-ds/datasource.ts
```diff
@@ -441,10 +441,11 @@ export class FlowDatasource {
     }
 
     if (combineIngressEgress) {
+      const ingressColumns =  flowSeries.columns.filter(column => column.ingress);
+      const egressColumns =  flowSeries.columns.filter(column => !column.ingress);
+      const columns = ingressColumns.length >= egressColumns.length ? ingressColumns : egressColumns;
```
fooker, line 446: Better to use a unique combination of both?

aramosvizcarra: What I noticed was that ingress/egress flows are not always even in quantity. And when using the combineIngressEgress it was inmediatelly filtering the result by using ingress as the # of columns. 

If the # of columns has only egress flows or more egress than ingress then the result  will only consider ingress to iterate.

I was pulling some data from the demo site and I notice the flows for example returned are only ingress. 
So if it was the opossite scenario (only egress flows) then it would not be any data to display. This way I'm using the largest number of columns to decide how iterate over all the possible values and do the calculation.

aramosvizcarra: > Better to use a unique combination of both?

What I noticed was that ingress/egress flows are not always even in quantity. And when using the combineIngressEgress it was inmediatelly filtering the result by using ingress as the # of columns.

If the # of columns has only egress flows or more egress than ingress then the result will only consider ingress to iterate.

I was pulling some data from the demo site and I notice the flows for example returned are only ingress.
So if it was the opossite scenario (only egress flows) then it would not be any data to display. This way I'm using the largest number of columns to decide how iterate over all the possible values and do the calculation.

aramosvizcarra: I tacked this separately since Will was having same results. Insetad of filtering by columns by ingres/egress I just created a Set of unique columns from the series, avoiding any possible mismatching scenario.

RangerRick (approved):
aha, makes sense

PR #422, opened 2022-05-24
HELM-317: Grafana Datasource expressions - Flows

aramosvizcarra:
Based on las feedback from jwood and Will "_I think best practice would be ensuring that the filter queries in flowds work seamlessly with multi-value template variables in Grafana. For example, if I were to create a template variable for flow classified Applications, and enable the Multi-Value option, and select 2 or more applications, I'd expect the withApplication($applications) filter to filter flows by the selected applications_"

Added 3 more template variables to allow multiple selection by applications, hosts and conversations which was already  partially supported by Helm. I also ensured that it will work with template variabales as well as constant values.

This template variables can be used with similar parameters as their corresponding rest apis.

- flows/applications/enumerate
- flows/conversations/enumerate
- flows/hosts/enumerate

Documentation on this ticket and how to use the template variables can be found  for the docs [PR]( https://github.com/OpenNMS/opennms-helm/pull/424/files)

I was hoping to be able to map host values to labels and conversation values to labels as well since they currently only return values and they are not really user friendly.

This instead of using expressions it allow to use template variables or fixed values for applications, hosts and conversations in combination with their corresponding withApplications(), withHosts() and withConversation() filter functions

Using template variables
Screen Shot 2022-05-24 at 12 13 09 PM

Template Variables (currently allow result limit value only as a parameter) 
Screen Shot 2022-05-24 at 12 14 54 PM

Using constant values
Screen Shot 2022-05-24 at 12 16 30 PM

Using template variables
Screen Shot 2022-05-24 at 12 18 34 PM

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-317


aramosvizcarra:
> the conversations display in the pulldown is kind of ugly, but functional, so otherwise this looks reasonable to me
Agreed, I was hoping to be able to map them to something but couldn't think of anything. I can create a jira to look into that later.

PR #453, opened 2022-06-29
HELM-272: Fix issues found while fixing HELM-272

synqotik:
HELM-272: Fix issues found while fixing HELM-272 which were causing FilterPanel not to update correctly

This needs to be done in conjunction with OpenNMS/opennms HELM-272, [PR](https://github.com/OpenNMS/opennms/pull/4958)

- In `lib/utils.ts`, `variableUpdated`, sometimes variables don't have `updateOptions` method defined, so check first
- In 'lib/utils.ts', added `getValueOrFirstElement()` since sometimes Grafana returns values that are either the raw value or else a one-element array with the value
- In `lib/text_box_variable.ts`, use `getValueOrFirstElement` to prevent `.trim() not found` error 

These issues were preventing Filter value dropdowns/text boxes to appear on Filter Panel

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-272

### DIFF
--- a/src/lib/text_box_variable.ts
+++ b/src/lib/text_box_variable.ts
@@ -1,4 +1,4 @@
-import { assignModelProperties, setOptionAsCurrent, setOptionFromUrl } from './utils';
+import { assignModelProperties, getValueOrFirstElement, setOptionAsCurrent, setOptionFromUrl } from './utils';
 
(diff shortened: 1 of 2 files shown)


aramosvizcarra (approved):
Looks good to me!

PR #570, opened 2022-11-29
HELM-354: Entity DS React updates, get metricFindQuery and template variables working

synqotik:
HELM-354: Entity DS React updates, get metricFindQuery and template variables mostly working

Will need more work on this, and looking to get feedback, but this PR does the following:

- Get `metricFindQuery` at least mostly working, particularly for Node queries
- Template variable substitution for Node queries
- Handle some attribute mapping
- Added `OnmsColumn` and `OnmsTableData` to make return values more strongly typed
- Add some more typing for return values, etc.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-354


scthompsonopen (approved):
Scott and I had a quick conversation, and we're going to implement some feedback in upcoming PRs because it's more style/convention than specific feedback or issues with code (which looks great).

Just for posterity, here's the feedback/style decisions we landed on together:
1. Keep Datasource classes lean as possible by moving supporting functions to helper/service files.
2. Try to avoid coding directly with Promises (even in transferred code from angular-based datasources), instead favouring async/await and Typescript's built-in functionality for automatically resolving synchronous responses.
3. Single return statements from functions/switch-case over multiple return statements where possible.

PR #573, opened 2022-11-29
HELM-354: Entity DS React fixes as per PR review suggestions

synqotik:
Various fixes to React Entity Datasource as per previous PR review suggestions.

- return once at end of `switch/case` statements
- move some helper functions from `EntityDatasource` into `EntityHelpers`
- remove explicit `Promise` statements from `async` functions
- fix `templateSrv.variables` -> `templateSrv.getVariables()` (`.variables` was deprecated), but only in new datasource
- a couple other small fixes

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/HELM-354


aramosvizcarra (approved):
They look good to me.
So are we ok with the not using semicolon?

scthompsonopen (approved):
Super approved. Awesome.

PR #595, opened 2023-01-18
HELM-360: Wire up React Filter Panel with Entity DataSource

synqotik:
HELM-360: Wire up React Filter Panel with Entity DataSource plus other updates.

## Overview

Connect the React FilterPanel to the React Entity DataSource, both for `metricFindQuery` queries for getting Filter Panel properties, as well as applying the Filter Panel selections to Entity DataSource queries.

While Grafana template variables can be used, the Filter Panel also allows filtering to be done over all panels in a dashboard without having to first specify template queries.

- Filter Panel selections are saved in browser Local Storage for use by the Entity DataSource

- changes made in Filter Panel Options editor are immediately reflected in the Filter Panel and most data is retained in local storage

- note, user needs to click the Dashboard Refresh button before Entity DataSource is re-queried and other panels update

- Filter Panel filters are applied in Entity DataSource per entity query, filters are added after any others. Handles both single and multi filter property selection

- changed Filter Panel filters to be displayed in rows instead of one long column. Filter Panel is scrollable and dropdowns display over any other adjacent panels

- user can select between Featured Attributes being displayed or not in the Filter Panel Options editor

- there were issues syncing filter order between Options and Control panels, and for example syncing metric values with the associated filter, by just using array indices, so needed to use a filterId (entity name + attribute) as a unique identifier. Also, metric data (i.e. sample values for filter items) is only saved in the Filter Panel itself and not in local storage or Options data

## Additional notes or items for future consideration:

- Filter Panel data, including filter definitions and actual filter values selected, is saved in browser local storage. Then, the React Entity DataSource checks for it before any query. Assuming this won't be a performance issue.

- Local storage could be retained even after any Filter Panels have been deleted. Should consider both Filter Panel having a hook to delete data when panel is deleted; and/or Entity DS should check for existence of a Filter Panel on current Dashboard

- May need to remove any sensitive info (credentials, etc.) from Entity DS data saved in local storage. May require then retrieving and possibly caching it

- Local storage may be removed during browser refresh or timeouts, user may lose actual Filter Panel value selections in those cases.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-360

### DIFF
--- a/src/datasources/entity-ds-react/EntityDataSource.ts
+++ b/src/datasources/entity-ds-react/EntityDataSource.ts
@@ -18,8 +18,9 @@ import {
 import { ClientDelegate } from 'lib/client_delegate'
 import { SimpleOpenNMSRequest, getNodeFilterMap } from 'lib/utils'
 import { getAttributeMapping } from './queries/attributeMappings'
-import { buildQueryFilter } from './queries/queryBuilder'
+import { buildQueryFilter, mergeFilterPanelFilters } from './queries/queryBuilder'
 import { EntityDataSourceOptions, EntityQuery, EntityQueryRequest, OnmsTableData } from './types'
+import { loadFilterEditorData } from '../../lib/localStorageService'
 
 export interface OnmsQueryResultMeta extends QueryResultMeta {
     entity_metadata: any[]
@@ -44,9 +45,26 @@ export class EntityDataSource extends DataSourceApi<EntityQuery> {
     async query(request: EntityQueryRequest<EntityQuery>): Promise<DataQueryResponse> {
         const fullData: OnmsTableData[] = []
 
+        // TODO:
+        // - May need to get dashboard info (via backendSrv or direct via Dashboard HTTP API)
+        //   and see if there is actually a FilterPanelReact in the current dashboard.
+        //   Right now just checking for localStorage data, which could
+        //   possibly exist even if the FilterPanelReact had been deleted
+        // - FilterPanelReact should perhaps get events to know when it has been deleted and clear
+        //   localStorage. May be able to subscribe to EventBus
+
+        // get data from any FilterPanels (React-only) that may be active
+        const filterEditorData = loadFilterEditorData()
+        const hasFilterEditorData = (filterEditorData &&
+            filterEditorData?.activeFilters.length > 0 && filterEditorData?.selectableValues.length > 0)
+
         for (let target of request.targets) {
             const entityType = target?.selectType?.label || EntityTypes.Alarms
             const filter = buildQueryFilter(target?.filter || new API.Filter(), request, this.templateSrv)
+
+            if (hasFilterEditorData) {
+                mergeFilterPanelFilters(entityType, filter, filterEditorData)
+            }
 
             try {
                 const rowData = await queryEntity(entityType, filter, this.client)
@@ -60,6 +78,16 @@ export class EntityDataSource extends DataSourceApi<EntityQuery> {
     }
 
     async metricFindQuery(query, options) {
+        // TODO: Might be good to be able to pass in options.entityType + options.attribute
+        // Otherwise callers may have to construct a fake query like 'nodes(label)' which we
+        // have to parse here, instead of passing in say:
+        /*
+            const options = {
+                entityType: 'node',
+                attribute: 'label'
+            }
+        */
+
         if (isLocationQuery(query)) {
             return metricFindLocations(this.simpleRequest)
         }
--- a/src/datasources/entity-ds-react/queries/queryBuilder.ts
+++ b/src/datasources/entity-ds-react/queries/queryBuilder.ts
@@ -1,7 +1,10 @@
 import { TemplateSrv } from '@grafana/runtime'
 import { API } from 'opennms'
+import { EntityQuery, EntityQueryRequest } from './../types'
+import { getAttributeMapping } from './attributeMappings'
 import { EntityTypes } from '../../../constants/constants'
-import { EntityQuery, EntityQueryRequest } from '../types'
+import { getFilterId } from './../../../panels/filter-panel-react/FilterPanelHelper'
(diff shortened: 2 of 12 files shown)


src/datasources/entity-ds-react/queries/queryBuilder.ts
```diff
@@ -1,7 +1,10 @@
 import { TemplateSrv } from '@grafana/runtime'
 import { API } from 'opennms'
+import { EntityQuery, EntityQueryRequest } from './../types'
+import { getAttributeMapping } from './attributeMappings'
 import { EntityTypes } from '../../../constants/constants'
-import { EntityQuery, EntityQueryRequest } from '../types'
+import { getFilterId } from './../../../panels/filter-panel-react/FilterPanelHelper'
```
aramosvizcarra, line 6: I would recommend to maybe move this functions (FilterEditorData and getFilterId) under a helper class in this directory (datasource) or in a common directory (lib) so they are not depending on a different component.

aramosvizcarra (changes requested):
Just one suggestion. Looks good in general. Unfortunately I wasn't able to load the panel. I've been trying to compile the whole plugin but I still can't see it listed in grafana.  I'll keep trying to I can do a better test.

src/datasources/entity-ds-react/queries/queryBuilder.ts
synqotik: Good recommendation!

I moved data structures mostly into Entity DS, and then moved functions and data structure access into `hooks/useFilterData`.

Now Entity DS has no dependencies on FilterPanel, and FilterPanel can access data and functions from this hook instead of digging into the EntityDS itself.

src/datasources/entity-ds-react/EntityDataSource.ts
```diff
@@ -44,10 +45,27 @@ export class EntityDataSource extends DataSourceApi<EntityQuery> {
     async query(request: EntityQueryRequest<EntityQuery>): Promise<DataQueryResponse> {
         const fullData: OnmsTableData[] = []
 
+        // TODO:
```
scthompsonopen, line 48: Yeah this solution is definitely not ideal, the use of localStorage, but this whole filter panel is a hack to begin with, so having to use further hacks to make it work only makes sense.

Another option I thought about a week or so ago for this problem, would be to store the relevant JSON data as a JSON.stringify() string within the DOM. Maybe with an ID so it's easy to retrieve. Then we can just parse that JSON string to get the data, and it would ensure that the Filter panel data is always tied to the right dashboard, and that the data goes away when deleting the panel, AND it means that a user can wipe their localstorage and still get their filter panel back.

Just a thought. :)

src/datasources/entity-ds-react/EntityDataSource.ts
```diff
@@ -60,6 +78,16 @@ export class EntityDataSource extends DataSourceApi<EntityQuery> {
     }
 
     async metricFindQuery(query, options) {
+        // TODO: Might be good to be able to pass in options.entityType + options.attribute
```
scthompsonopen, line 81: I'm a fan of capturing TODOs in tickets only so that we don't forget. So easy to lose TODOs like this otherwise.

src/datasources/entity-ds-react/queries/queryBuilder.ts
```diff
@@ -171,3 +173,41 @@ export const buildQueryFilter = (filter: API.Filter, request: EntityQueryRequest
...
+
+    for (const filter of filtersToMerge) {
+        const filterId = getFilterId(filter)
+        const selectableValues = filterEditorData?.selectableValues.find(sv => sv.filterId === filterId)
+
+        if (selectableValues?.values) {
+            // get all selected values that are non-empty, converted to strings
+            const values = selectableValues?.values.map(sv => (sv.value || '').toString()).filter(v => v.length > 0)
+
+            if (values) {
```
scthompsonopen, line 193: Nit: We could probably break this function down into a few chunks. 4 levels of indentation feels like a lot.

src/panels/filter-panel-react/FilterPanelControl.tsx
```diff
@@ -1,45 +1,214 @@
...
+    useEffect(() => {
+        const filterEditorData = loadFilterEditorData()
+
+        if (filterEditorData && filterEditorData.selectableValues) {
+            setSelectableValues(filterEditorData.selectableValues)
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [])
+
+    useEffect(() => {
```
scthompsonopen, line 30: This useEffect and the getValuesFromDatasource below feel like they could be broken down into smaller chunks.

src/panels/filter-panel-react/FilterPanelControl.tsx
```diff
@@ -1,45 +1,214 @@
...
+        saveFilterPanel(newValues)
+    }
+
+    /**
+     * Save filter panel data to localStorage, for use by Entity Datasource when filtering queries.
+     */
+    const saveFilterPanel = (filterSelectableValues: FilterSelectableValues[]) => {
         if (props.options.filterEditor) {
-            saveFilterPanel(props.options.filterEditor)
+            // TODO: May want to remove some data from datasource before saving
```
scthompsonopen, line 173: This should definitely be tracked in a JIRA and double checked before going live, this was a quick and dirty solution and your concerns about possibly storing sensitive data is legit and should be verified.

scthompsonopen (approved):
Approving for the deadline and because my comments are mostly nits, and the ones that weren't were comments on the lengths of functions. All stuff that can be tackled in the future.

aramosvizcarra (approved):
Looks pretty good!

synqotik:
Merging this. Will capture issues raised in the various comments in separate Jira tickets to tackle in subsequent work.

PR #625, opened 2023-03-10
HELM-373: Dashboard conversion tool, initial version

synqotik:
HELM-373. Dashboard conversion tool.

This is a Helm / Grafana panel in which the user would paste the Dashboard Json from an existing Helm 8 or previous dashboard, click Convert and the converted Json will appear. User can then import that dashboard json.

User will need to have saved off previous json, then install Helm 9 and the new datasources for this tool to work.

This will fix up any datasource references to use the Helm 9 / React datasources and attempt to convert data structures to the new Helm 9 / React schemas.

This is a work in progress but want to get it reviewed and into the codebase so we can all update and improve it.

- Added "version" field to all plugin metadata so we can distinguish Helm 9+ datasource and panel plugins
- Performance DS Attribute, Expression and Filter queries are best supported
- Performance String Property queries being worked on (subsequent PR)
- Entity queries are partially supported, still working on these. Still issues with complex/nested queries
- Flow queries need to be done
- Does not convert any Helm panels, needs to be done
- Added "Unhide all queries" switch, this will set `hide` to `false` for all targets. I found this to be helpful as some existing dashboards seemed to have had some targets hidden, which caused issues for say Expression queries which tried to reference a previous, hidden attribute
- Checks `__inputs` and `templating` (template variables) sections for datasource references and converts the name and uid to the Helm 9 versions
- Currently users would still need to edit the Dashboard json before importing, for example changing the dashboard UID or title. We should add this into the tool
- May want someone in UX to review and offer suggestions

Screenshot 2023-03-10 at 11 48 11

# External References


scthompsonopen (approved):
LGTM!

PR #629, opened 2023-03-14
HELM-376: Grafana 9 Migration from grafana-toolkit

synqotik:
Use Grafana 9 migration tool to migrate from `grafana-toolkit`. Also update to Grafana 9 runtime libraries, React App component.

This PR attempts to do the minimal required to run without actually removing all the legacy Angular plugins. However, note that the Angular plugins won't actually work since parts of the code have been removed or commented-out.

Once this builds in CircleCI and runs, we will remove the Angular code and do more cleanup. Some of this code is intentionally incomplete or messy.

- ran Grafana plugin migration tool `npx @grafana/create-plugin@latest migrate`
- updated all Grafana libraries to 9.3.8
- fixes to CircleCI build commands, may need to be further updated in subsequent PRs. Also, signing fails, may need to wait for main branch merge to see if that and subsequent build steps work
- ported Alberto's work on `App.tsx` component
- fixed bunch of lint issues (mostly replacing `var` with `let` or `const`)
- only run React jest tests, to avoid context setup issues with Angular
- fixups like removing jquery plot/flot. Angular plugins are *not* expected to work, we are just not quite removing them yet
- left React plugins with `-react` in their names, this may be changed later
- new build process uses `yarn` however left `package-lock.json` in since CircleCI uses it in the Node Package Checksum step; will eventually need to fix

# External References

* JIRA (Issue Tracker): [HELM-376](https://opennms.atlassian.net/browse/HELM-376)


scthompsonopen (approved):
I'm approving this but I don't have much to comment on as it's mostly auto-generated by the tool. What's in here makes sense from a modern React perspective, and the eslint rules forcing the var -> const/let update makes sense as well.

.config/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+ * https://grafana.github.io/plugin-tools/docs/advanced-configuration#extending-the-typescript-config
+ */
+ {
+  "compilerOptions": {
+    "alwaysStrict": true,
+    "declaration": false,
+    "rootDir": "../src",
+    "baseUrl": "../src",
+    "typeRoots": ["../node_modules/@types"],
+    "resolveJsonModule": true
```
aramosvizcarra, line 14: I had to add this line to avoid building errors
```suggestion
    "resolveJsonModule": true,
    "noImplicitAny": false
```

tsconfig.json
```diff
@@ -1,10 +1,6 @@
 {
-  "extends": "./node_modules/@grafana/toolkit/src/config/tsconfig.plugin.json",
-  "include": ["src", "types"],
+  "extends": "./.config/tsconfig.json",
   "compilerOptions": {
-    "rootDir": "./src",
-    "baseUrl": "./src",
-    "noImplicitAny": false,
-    "typeRoots": ["./node_modules/@types"]
+    "noImplicitAny": false
```
aramosvizcarra, line 4: aha!

aramosvizcarra (changes requested):
I think yarn.lock is not necessary. Is not in grafana instructions but we can still use npm instead of yarn.

Another thing I didn't see is the updated plugin.json updated [here](https://grafana.com/docs/grafana/v9.0/developers/plugins/metadata/#includes) they changes what properties are required. There are a few more required than the current ones for `info`

aramosvizcarra (approved):

synqotik:
> I think yarn.lock is not necessary. Is not in grafana instructions but we can still use npm instead of yarn. (not really important)

good, yes let's follow up and decide on one or the other. I will test it out after this PR.

> Another thing I didn't see is the updated plugin.json updated [here](https://grafana.com/docs/grafana/v9.0/developers/plugins/metadata/#includes) they changes what properties are required. There are a few more required than the current ones for [info](https://grafana.com/docs/grafana/v9.0/developers/plugins/metadata/#info) This can be done after when we complete the cleanup.

It looks like we already have those required `info` fields in the main `plugin.json`, so they didn't show up in the diff. May need to add them to the individual plugins.

synqotik:
Merging to get these changes in, we will need to fix up some things afterwards. Also to see whether the signing step works in the main feature branch; may need to make that optional.

PR #652, opened 2023-03-27
HELM-373: Refactor dashboard convert, filter panel convert

synqotik:
Refactor dashboard conversion code into a module under 'lib', and break out into separate files.

Add Filter Panel conversion.

Remove dashboard UID, so user is prompted to enter another (or have Grafana auto-generate it), to avoid any ID conflicts.

In Dashboard Convert Panel, enable setting dashboard title.

As of now, any selected values previously set in the legacy Filter Panel will be lost, since the new Filter Panel saves these using `localStorage` instead of in dashboard json. May add that in, but we should probably add some code to save/restore Filter Panel data via `localStorage` by dashboard ID first.

# External References


aramosvizcarra (approved):
Looks very good!

PR #666, opened 2023-04-03
Fix Performance DS Filter Query, add StringProperty conversion

synqotik:
Various Performance datasource fixes:

- in Performance Filter query editor, initialize from query props
- in Perf Filter query editor, use `key` instead of `displayName` as `filter` parameter and `filterState` object key
- fix to Attribute Query `fallbackAttribute` conversion
- conversion tool implementation of Performance `stringProperty` query. Needs work but better than previous

Need some more work for String Property conversion. For example the legacy `nodeId` may be in `FS:FI` format, but new React version looking for node id and/or host label; and legacy `resourceId` may not match React version. May need to do API lookups to fill in values correctly. However, at least the user can see the legacy values after dashboard conversion and make updates.

# External References

### DIFF
--- a/src/datasources/perf-ds/PerformanceFilter.tsx
+++ b/src/datasources/perf-ds/PerformanceFilter.tsx
@@ -25,8 +25,8 @@ export interface PerformanceFilterProps {
 
 export const PerformanceFilter: React.FC<PerformanceFilterProps> = ({ query, updateQuery, loadFilters }) => {
     // TODO: Initialize these from 'query' prop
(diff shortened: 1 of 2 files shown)


aramosvizcarra (approved):
LGTM. 
Although sonar cloud seemd to have found a bug. Otherwise I'm good with it.

scthompsonopen (dismissed):

synqotik:
> LGTM. Although sonar cloud seemd to have found a bug. Otherwise I'm good with it.

fixed the issue and rebased

aramosvizcarra (approved):

PR #673, opened 2023-04-11
HELM-395: Fix issue with Alarm Table Panel 'freezing'

synqotik:
The Alarm Table Panel appeared to 'freeze' after clicking it a few times. This was caused by some issues with how event handlers were being added.

This PR fixes that issue, but note if you for example change the data range using the Grafana controls, you will get new data, but event handlers won't be attached and you need to refresh the browser window to get the click / context menu functionality back. Will need to fix this.

This also pushes an updated `package-lock.json` to make sure the plugin names are in sync with `package.json`.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-395

### DIFF
--- a/src/panels/alarm-table/hooks/useAlarmTableMenu.ts
+++ b/src/panels/alarm-table/hooks/useAlarmTableMenu.ts
@@ -3,13 +3,13 @@ import { useState, useEffect, useRef } from 'react'
 export const useAlarmTableMenu = (rowClicked) => {
 
     const [menu, setMenu] = useState({ x: 0, y: 0 })
-    const [menuOpen, setMenuOpen] = useState(false);
+    const [menuOpen, setMenuOpen] = useState(false)
 
     const contextMenu = (index: number, e: MouseEvent) => {
-        e.preventDefault();
-        rowClicked(index, e,true);
+        e.preventDefault()
+        rowClicked(index, e, true)
         setMenu({ x: e.x, y: e.y })
-        setMenuOpen(() => true);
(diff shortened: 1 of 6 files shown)


src/panels/alarm-table/hooks/useAlarmTableMenu.ts
```diff
@@ -19,16 +19,45 @@ export const useAlarmTableMenu = (rowClicked) => {
...
+        const onRowClicked = (e: Event) => processRowClicked(e)
+        const onContextMenu = (e: Event) => processContextMenu(e)
+
         rows?.forEach((row, index) => {
-            row.removeEventListener('click', (e: unknown) => rowClicked(index, e as MouseEvent))
-            row.removeEventListener('contextmenu', (e: unknown) => contextMenu(index, e as MouseEvent))
-            row.addEventListener('click', (e: unknown) => rowClicked(index, e as MouseEvent))
-            row.addEventListener('contextmenu', (e: unknown) => contextMenu(index, e as MouseEvent))
+            row.setAttribute('data-rowindex', '' + index)
+            row.addEventListener('click', onRowClicked)
```
scthompsonopen, line 48: To fix this permanently, we'd just have to attach the event listener to an element that isn't removed from the DOM (a parent), or add a few more variable changes to the useEffect callback. You mentioned in the PR description that when you change the date range, the listeners aren't re-attached, so we might just have to pull in the date values into the [table, menu, menuOpen, rowClicked] variables.

PR #684, opened 2023-04-20
HELM-408: Dashboard Conversion and Alarm Table Fixes

synqotik:
Fixes to dashboard conversion code plus some Alarm Table fixes.

- added back `helm` to conversion code that looks for legacy datasources and panels (this had been inadvertently removed)
- force Alarm Table conversion for scrolling to be true (more often that not, horizontal scroll will be needed to display all columns)
- fix Alarm Table number conversion, it's actually `none` instead of `number`
- handle case where legacy datasource has `uid` of a template variable but no `type`
- fix null reference issue

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-408


aramosvizcarra (approved):
looks great!

PR #685, opened 2023-04-21
run `release-lint.sh` on `main-*` or tags

RangerRick:
Run the `release-lint.sh` script (which checks for `-SNAPSHOT` versions and `npm audit` errors)


mershad-manesh (approved):

RangerRick:
It actually runs on _every_ run, but is only fatal on "release" tags. :)

PR #686, opened 2023-04-21
HELM-398: Enable 'Use Grafana user' in Alarm Table Panel

synqotik:
Enable a "Use Grafana user" switch for the Alarm Table Panel.

It's set in the Alarm Table Options panel. If set, it will send the username of the currently logged-in Grafana user as the user for alarm ack and other actions. Otherwise, the user is left "unset", in which case the user that is configured in the Entity Datasource will be used.

In Helm 8, the control for this was in the Entity Datasource options; however since we actually call the OpenNMS alarm API from Alarm Table Panel code, I put the "Use Grafana user" switch in the Alarm Table Panel options.

Also as a note, we use the Grafana "Actual user" API which returns info about the current logged-in Grafana user. This includes an `id` which is an integer value assigned by Grafana that has nothing to do with OpenNMS, and a `login` string field, which should correspond to an OpenNMS username; we use `login`.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-398

### DIFF
--- a/src/panels/alarm-table/AlarmTableAdditional.tsx
+++ b/src/panels/alarm-table/AlarmTableAdditional.tsx
@@ -1,0 +1,48 @@
+import React, { useEffect, useState } from 'react'
+import {
+  InlineField,
+  InlineFieldRow,
+  Switch
+} from '@grafana/ui'
+import { AlarmTableAdditionalState } from './AlarmTableTypes'
+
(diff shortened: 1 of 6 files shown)


aramosvizcarra (approved):
Looks good!

scthompsonopen (approved):

PR #703, opened 2023-05-11
update XLSX to latest version

RangerRick:
they are no longer updating npm, just distributing through another CDN


aramosvizcarra (approved):
Yup

PR #709, opened 2023-05-22
HELM-424, HELM-425: Fix transient flag logic in Perf DS queries, node selection

synqotik:
HELM-425: Fix logic for sending `transient: true` in Performance data source Measurement API queries. Should only be sent when a query has been flagged as `hidden`.

Previous logic resulted in `transient: true` almost always being sent, resulting in `204: No Content` responses, making it appear that Performance data source was not working at all.

HELM-424: Fix some issues which caused user to not be able to select a node from the Node dropdown in a Performance query.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-424,
* https://opennms.atlassian.net/browse/HELM-425

### DIFF
--- a/src/datasources/perf-ds/queries/queryBuilder.ts
+++ b/src/datasources/perf-ds/queries/queryBuilder.ts
@@ -44,10 +44,10 @@ export const isValidMeasurementQuery = (query: OnmsMeasurementsQueryRequest) => 
 
 export const isValidAttributeTarget = (target: PerformanceQuery) => {
     if (!target ||
-        !(target.attribute &&
-          (target.attribute.attribute.name || (target.attribute.attribute.label && OpenNMSGlob.hasGlob(target.attribute.attribute.label))) &&
-          (target.attribute.resource.id || target.attribute.resource.label) &&
-          (target.attribute.node.id || target.attribute.node.label))) {
+        !target.attribute ||
+        !((target.attribute.attribute?.name || (target.attribute.attribute?.label && OpenNMSGlob.hasGlob(target.attribute.attribute?.label))) &&
+          (target.attribute.resource?.id || target.attribute.resource?.label) &&
+          (target.attribute.node?.id || target.attribute.node?.label))) {
         return false
     }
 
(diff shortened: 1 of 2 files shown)


mmahacek:
In my test, it appears that the transient status is being set properly based on if the query is set as hidden in the editor.

RangerRick (dismissed):
ooooooooh, sneaky logic bug :D

nice catch

synqotik:
> ooooooooh, sneaky logic bug :D
> 
> nice catch

quad-state `boolean` heh

src/datasources/perf-ds/queries/queryBuilder.ts
```diff
@@ -44,10 +44,10 @@ export const isValidMeasurementQuery = (query: OnmsMeasurementsQueryRequest) =>
...
-        !(target.attribute &&
-          (target.attribute.attribute.name || (target.attribute.attribute.label && OpenNMSGlob.hasGlob(target.attribute.attribute.label))) &&
-          (target.attribute.resource.id || target.attribute.resource.label) &&
-          (target.attribute.node.id || target.attribute.node.label))) {
+        !target.attribute ||
+        !((target.attribute.attribute?.name || (target.attribute.attribute?.label && OpenNMSGlob.hasGlob(target.attribute.attribute?.label))) &&
+          (target.attribute.resource?.id || target.attribute.resource?.label) &&
+          (target.attribute.node?.id || target.attribute.node?.label))) {
         return false
     }
```
RangerRick, line 52: I wonder if it's worth breaking out this stuff into `const`s and having a simpler chain, just to make it a bit more readable. I would hope the compiler would optimize it away for actual runtime anyway.

RangerRick (approved):
readability aside, seems like a reasonable way to handle it

src/datasources/perf-ds/queries/queryBuilder.ts
synqotik: will refactor next time we touch this file

PR #710, opened 2023-05-26
Plugin and Docker Validation

RangerRick:
This PR adds plugin and docker image trivy validation to the grafana plugin workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -275,13 +275,15 @@ jobs:
               -t grafana-plugin:$(.circleci/get-version.sh) .
       - run:
           name: Export Docker Image as Artifact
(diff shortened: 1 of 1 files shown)


jeffgdotorg (commented):
This looks great in terms of what it does. I'll leave the decision to approve the how to someone who fully groks that aspect.

mershad-manesh (approved):
Changes look good.

Would it be better if we create the artifact folder in the home directory (~/) ?

RangerRick:
> Changes look good.
> 
> Would it be better if we create the artifact folder in the home directory (~/) ?

The workflow is using the workspace to pass the artifacts around for everything else, just not the oci. Figured I'd make it match, wasn't gonna completely re-engineer to download them or anything.

PR #719, opened 2023-06-07
HELM-414: Fix PerfDS label conversion, add 'requires' conversion

synqotik:
Fix Performance data source converter to move Attribute `label` correctly. This was causing Performance queries with Expressions to fail after conversion.

Also adds data source conversion for `__requires` section.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-414


scthompsonopen (approved):
LGTM!

PR #720, opened 2023-06-07
HELM-422: Fix Dashboard Import for included dashboards

synqotik:
Importing the included dashboards via "Configuration > Plugins > Dashboards" was failing.

Fix includes:

- setting dashboard `id` to null, looks like this is needed to import correctly
- removing `uid`
- `version` needs to be same or 1 more than previous value. Grafana does a check if version is mismatched, thinking it means "someone else" updated it. It could be that having a gap in version triggers this error.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-422


mmahacek (approved):
🕺

scthompsonopen (approved):

PR #721, opened 2023-06-08
HELM-396: Click anywhere in Alarm Table for context menu

synqotik:
Allow user to right-click (context click, could be Ctrl-click or double-tap on Mac) anywhere in the background area of the Alarm Table (not just on a row) to open the context menu.

Left-click results in deselecting any previously-selected rows.

Right-click (context click) will select new item and display context menu for that item, if there previously had been 0 or 1 items selected. If there were 2+ items selected, it will display context menu and not select item under it.

If there are multiple items selected, context menu displays number of items next to action, e.g. "Acknowledge (2)", etc.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-396

### DIFF
--- a/src/panels/alarm-table/AlarmTableControl.tsx
+++ b/src/panels/alarm-table/AlarmTableControl.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useRef } from 'react'
 import { PanelProps } from '@grafana/data'
 import { Button, ContextMenu, Modal, Pagination, Tab, TabContent, Table, TabsBar } from '@grafana/ui'
 import { AlarmTableMenu } from './AlarmTableMenu'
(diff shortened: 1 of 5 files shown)


RangerRick (approved):
I don't know react but I _think_ I understand what you're doing here. Seems reasonable.

synqotik:
> This doesn't work on my test. I have two alarms, and if I left click one and right click a different one, the Details that are show are for the one that has been left-clicked to select as the current row, not for the alarm under the mouse pointer.

This should now be fixed. See description for how behavior works, should now be the same as legacy panel.

RangerRick (commented):
I will defer to @mmahacek who has an actual ability to test this easily :D

mmahacek (approved):
Clicking on a non-selected row is moving the focus for me so the expected row is actioned in the context menu.

PR #748, opened 2023-06-15
HELM-435 Fix multi-valued template variables in Entity queries

synqotik:
Multi-valued template variables were not being processed correctly resulting in them not working and sending out incorrect FIQL, for example when used in the Alarm Table panel.

Looks like Grafana 9 uses `multi` instead of `isMulti` for multi-valued template variables, so we were not catching them and converting into nested restrictions. Also if a multi-valued variable was found, it was skipping processing subsequent clauses, like the date range clause.

Also:

- added some `min-width` for Entity query clause variable values so it's easier to enter a template variable
- fixed a small issue in Flows query

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-435

### DIFF
--- a/src/datasources/entity-ds/EntityClause.tsx
+++ b/src/datasources/entity-ds/EntityClause.tsx
@@ -88,6 +88,14 @@ export const EntityClause = ({
     }
 
     return (
+        <>
+        <style>
+          {`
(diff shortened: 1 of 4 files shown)


mmahacek (approved):
LGTM.  And WOMM.

PR #779, opened 2023-06-26
HELM-440: Fix date and multi-value handling in Entity clauses and queries

synqotik:
Fixed some issues with Entity data source date and multi-value handling, which were causing behavior in the bug ticket.

- use either `comparedString` or `comparedValue` in query clauses, but not both. This was causing some mixups with how queries were sent to the Rest API

- using `datetime-local` input type for dates, allows user to set the full date and time and uses ISO date format when sending to Rest API

- remove any empty query clauses before sending to `opennms-js` and doing the Rest query. This could happen when a mult-value template variable is used but all options are deselected, and was causing an invalid query to be sent. Behavior now is as if the filter was not applied at all

- also fixes HELM-442 issue entering manual text (like a template variable) in an Entity DS `where` clause

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-440

### DIFF
--- a/src/datasources/entity-ds/queries/queryBuilder.ts
+++ b/src/datasources/entity-ds/queries/queryBuilder.ts
@@ -161,13 +161,16 @@ export const getTemplateVariable = (templateSrv: any, name: string) => {
     return undefined
 }
 
+const isValidClauseRestriction = (clause) => {
+    return clause?.restriction?.attribute && clause?.restriction?.comparator && clause?.restriction?.value
+}
+
 // Given a query filter, build a cloned one which does template variable substitution, etc.
 // Clone Filter to make substitution possible
 // (otherwise substitution would happen in original query,
 // and overwriting the $<variable> or [[variable]] in restrictions which may not be the intention)
 export const buildQueryFilter = (filter: API.Filter, request: EntityQueryRequest<EntityQuery>, templateSrv: TemplateSrv): API.Filter => {
-
-    filter.clauses = filter.clauses.filter((clause, idx) => clause.restriction?.attribute && clause.restriction?.comparator && clause.restriction?.value)
+    filter.clauses = filter.clauses.filter(clause => isValidClauseRestriction(clause))
     const clonedFilter = API.Filter.fromJson(filter)
 
     // Before replacing any variables, add a global time range restriction (which is hidden to the user)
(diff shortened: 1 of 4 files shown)


RangerRick:
> * remove any empty query clauses before sending to `opennms-js` and doing the Rest query. This could happen when a mult-value template variable is used but all options are deselected, and was causing an invalid query to be sent. Behavior now is as if the filter was not applied at all

@synqotik Is this a thing we should handle more gracefully in `opennms-js` directly?

synqotik:
> > * remove any empty query clauses before sending to `opennms-js` and doing the Rest query. This could happen when a mult-value template variable is used but all options are deselected, and was causing an invalid query to be sent. Behavior now is as if the filter was not applied at all
> 
> @synqotik Is this a thing we should handle more gracefully in `opennms-js` directly?

@RangerRick possibly - I'm loth to update `opennms-js` unless we are going to make a bunch of fixes ...

RangerRick:
@synqotik yeah I don't mean to block this fix, just in general, seems like a useful thing to do "upstream" as it were

synqotik:
> @synqotik yeah I don't mean to block this fix, just in general, seems like a useful thing to do "upstream" as it were

@RangerRick agreed, I'll take a look and maybe make a Jira ticket for that :)

src/datasources/entity-ds/queries/queryBuilder.ts
```diff
@@ -161,13 +161,16 @@ export const getTemplateVariable = (templateSrv: any, name: string) => {
...
+}
+
 // Given a query filter, build a cloned one which does template variable substitution, etc.
 // Clone Filter to make substitution possible
 // (otherwise substitution would happen in original query,
 // and overwriting the $<variable> or [[variable]] in restrictions which may not be the intention)
 export const buildQueryFilter = (filter: API.Filter, request: EntityQueryRequest<EntityQuery>, templateSrv: TemplateSrv): API.Filter => {
-
-    filter.clauses = filter.clauses.filter((clause, idx) => clause.restriction?.attribute && clause.restriction?.comparator && clause.restriction?.value)
+    filter.clauses = filter.clauses.filter(clause => isValidClauseRestriction(clause))
```
scthompsonopen, line 173: I know I'm a little late to the approval game here, but I just wanted to call out this small refactor for making me happy.

PR #792, opened 2023-06-29
HELM-446: Convert Perf datasource NaN values to nulls

synqotik:
Convert any 'NaN' values received by the Performance data source to nulls. This was done in v8 of this plugin.

Was causing issues with for example the Stat Grafana visualization panel using the Calculation "Last *", which displays the last non-null value, however it would consider `"NaN"` to be a valid value and display it instead of a valid numeric value.

Flows has the `nanToZero` feature which may be more appropriate for flows data.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-446

### DIFF
--- a/src/datasources/perf-ds/PerformanceHelpers.ts
+++ b/src/datasources/perf-ds/PerformanceHelpers.ts
@@ -1,8 +1,8 @@
-import { ArrayVector, DataFrame, Field, FieldType } from "@grafana/data"
-import { getTemplateSrv } from "@grafana/runtime";
-import { FunctionFormatter } from "lib/function_formatter";
-import { trimChar, isString } from "lib/utils";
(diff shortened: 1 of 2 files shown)


mmahacek (approved):
Tested artifact and was able to replicate working behavior from v8 👍

synqotik:
> Tested artifact and was able to replicate working behavior from v8 👍 image

thanks! have to fix a Sonar issue, so may need re-approval...

synqotik:
Refactored `PerformanceHelpers.measurementResponseToDataFrame` to make SonarCloud happy

mmahacek (approved):

synqotik:
Noted a few variables that could be `const`, I'll include those in my next PR.

PR #818, opened 2023-07-19
HELM-438: Dashboard converter retains original datasource name/label/uid

synqotik:
HELM-438: Dashboard converter retains original datasource name/label/uid when possible

Dashboard converter tool will retain the original datasource name and label if present, and if the datasource uid reference is a template or input variable, it will retain it instead of using the uid of the installed v9 datasource.

This will help in cases where multiple instances of an OpenNMS datasource are used, for example if they point to different OpenNMS instances.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/HELM-438

### DIFF
--- a/src/lib/dashboard-convert/datasources.ts
+++ b/src/lib/dashboard-convert/datasources.ts
@@ -4,7 +4,7 @@ import { getDatasourceTypeFromPluginId } from './utils'
 import { isString } from '../utils'
 
 // Datasource info found in panel or target/query
-interface SourceDatasourceInfo {
+export interface SourceDatasourceInfo {
   isOpenNmsDatasource: boolean
   isTemplateVariable: boolean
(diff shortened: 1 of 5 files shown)


scthompsonopen (approved):
Huge fan of taking a bunch of repeated code, sticking it in a function, and then adding a big comment. Super tiny nit would be to put the commend on updateTargetDatasource into js doc format, but otherwise it looks awesome to me.

RangerRick (approved):

synqotik:
> Huge fan of taking a bunch of repeated code, sticking it in a function, and then adding a big comment. Super tiny nit would be to put the commend on updateTargetDatasource into js doc format, but otherwise it looks awesome to me.

Thanks! I'll add the js doc format in next PR.

PR #846, opened 2023-08-16
OPG-460: Add SNMP Primary ifindex and Primary IP Address to node table

synqotik:
OPG-460: Add SNMP Primary ifindex and Primary IP Address to node table results

These column values weren't included, causing Categories values to display under SNMP ifIndex column.

Pulled in the logic from v8 to get these values.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/OPG-460

### DIFF
--- a/src/datasources/entity-ds/queries/queryNodes.ts
+++ b/src/datasources/entity-ds/queries/queryNodes.ts
@@ -1,5 +1,5 @@
-import { ClientDelegate } from "lib/client_delegate";
-import { OnmsNode } from "opennms/src/model/OnmsNode";
+import { ClientDelegate } from 'lib/client_delegate'
+import { OnmsNode } from 'opennms/src/model/OnmsNode'
 import { API } from 'opennms'
 import { OnmsColumn, OnmsTableData } from '../types'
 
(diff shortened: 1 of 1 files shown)


src/datasources/entity-ds/queries/queryNodes.ts
```diff
@@ -64,7 +80,9 @@ export const queryNodes = async (client: ClientDelegate, filter: API.Filter): Pr
             node.netBiosDomain,
             node.operatingSystem,
             node.lastCapsdPoll,
-            node.categories ? node.categories.map(cat => cat.name) : undefined,
+            ifIndex || '',
+            ipAddress,
+            node.categories ? node.categories.map(cat => cat.name) : ''
```
RangerRick, line 85: Is it going to mess anything up that the ordering of these changed?

src/datasources/entity-ds/queries/queryNodes.ts
```diff
@@ -1,5 +1,5 @@
-import { ClientDelegate } from "lib/client_delegate";
-import { OnmsNode } from "opennms/src/model/OnmsNode";
+import { ClientDelegate } from 'lib/client_delegate'
+import { OnmsNode } from 'opennms/src/model/OnmsNode'
```
RangerRick, line 2: it would be nice if rando editor format cleanups were in a separate commit so it's easier to review the "real" changes

(or we should just do one big cleanup pass with prettier or something if we want to enforce different styles)

RangerRick (approved):
tentative approval, assuming changing the "column" order of the returning array is OK

src/datasources/entity-ds/queries/queryNodes.ts
synqotik: The returned rows should match the order of the `columns` above, so this was actually a bug since these items were in the columns expected but not in the rows, so `Categories` data was where `Primary SNMP ifIndex` was expected.

src/datasources/entity-ds/queries/queryNodes.ts
synqotik: Good point. Maybe format cleanups could be a separate commit. I've been updating files as I change them, haven't decided whether to do one big cleanup PR.

PR #870, opened 2023-09-08
OPG-454: Allow Flows query with multi-valued withHost

synqotik:
Allow 'withHost' Flows queries with multi-valued parameters for the 'host' value.

Needed to convert the `withHost` parameter from a string to an array, e.g. from `withHost: '{host1,host2}'` to `withHost: ['host1', 'host2']`

This was causing `opennms-js` Rest API call to be:

```
http://localhost:3000/api/datasources/proxy/uid/some-id/rest/flows/hosts/series?start=1694178697745&end=1694200297745&step=30989&host={0.0.0.0,0.0.0.1}
```

instead of

```
http://localhost:3000/api/datasources/proxy/uid/some-id/rest/flows/hosts/series?start=1694178697745&end=1694200297745&step=30989&host=0.0.0.0&host=0.0.0.1
```

Also fixed another issue getting interfaces if nodes was a multi-valued variable.

There are other places where we may not handle multi-valued variables correctly, may refactor a bit as we find those.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/OPG-454

### DIFF
--- a/src/datasources/flow-ds/helpers.ts
+++ b/src/datasources/flow-ds/helpers.ts
@@ -3,11 +3,13 @@ import { Model } from 'opennms'
 import { ClientDelegate } from 'lib/client_delegate'
 import { dscpLabel, dscpSelectOptions } from '../../lib/tos_helper'
 import {
-    swapColumns,
+    getMultiValues,
     getNodeFilterMap,
     getNumberOrDefault,
     getNodeAsResourceQuery,
+    isMultiValueString,
     isString,
+    swapColumns,
     valueOrDefault
 } from 'lib/utils'
 import { SimpleOpenNMSRequest } from 'lib/simpleRequest'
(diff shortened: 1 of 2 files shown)


RangerRick (approved):

mmahacek:
Tested with data from demo.opennms.com.  Verified that I was able to select a handful of hosts and return data.
If I selected `All` hosts (about 25 results), I would get the following, but that's not related to this issue.
```
Failed to execute query: java.lang.Exception: {"root_cause":[],"type":"search_phase_execution_exception","reason":"","phase":"fetch","grouped":true,"failed_shards":[],"caused_by":{"type":"too_many_buckets_exception","reason":"Trying to create too many buckets. Must be less than or equal to: [10000] but was [10001]. This limit can be set by changing the [search.max_buckets] cluster level setting.","max_buckets":10000}}
```

mmahacek (approved):

PR #910, opened 2023-10-16
OPG-464: Add fields to monitored services

synqotik:
2nd part of OPG-464, added `ipAddress`, `down`, `ipInterfaceId`, `nodeId`, `nodeLabel` to EntityDS, Monitored Services. These were waiting on updates to Horizon and to `opennms-js`.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-464

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "flot-axislabels": "https://github.com/j-white/flot-axislabels#master",
         "lodash.clonedeep": "^4.5.0",
         "lodash.escaperegexp": "^4.1.2",
-        "opennms": "^2.5.6",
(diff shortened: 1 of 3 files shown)


mikewrosey (approved):
lgtm

mmahacek:
The fields are appearing in the table.  I'm noticing that I can't filter on the `Down` state as there isn't an option for that.  Also, filtering on `Management Status` always fails as it attempts to filter the `status` field based on the full text and it should be converting to the enum value.  

This: `&_s=status%3D%3DN"`
not this: `&_s=status%3D%3DNot%2520Monitored"`

PR #930, opened 2023-11-30
OPG-470: Alarm Panel auto refresh after action

synqotik:
Adds an option to auto-refresh the Alarm Panel after performing an action such as `acknowledge`, `clear` or `escalate` on one or more alarms, so the user sees realtime updated data after performing the action (e.g. `Ack` field displays "true" instead of "false" after an `Acknowledge`).

The Grafana runtime does not appear to offer a way to refresh a panel or the dashboard programmatically, so here we "click" the "Refresh dashboard" button. Note, there could be issues if Grafana changes or removes this button in a future version (the `RefreshPicker` component).

Grafana React appears to have removed `getTimeSrv()` or any other access to the `TimeSrv` service, which includes `refreshTimeModel()` which could be used to refresh the query and panel. See [TimeSrv](https://github.com/grafana/grafana/blob/c3ebd388e3cc192590c3f4d381429f7d9e345765/public/app/features/dashboard/services/TimeSrv.ts#L250).

Using `PanelProps.onChangeTimeRange` kind of works, however it takes an `AbsoluteTimeRange` which contains `from` and `to` properties as timestamps. However, it does not contain the `raw` property of `TimeRange` which allows specifying things like `now` or `now-6M` ("Last 6 months"), so any call to `onChangeTimeRange` will replace that with an absolute time range, which would be undesirable from a user standpoint.

The Alarm Table Panel Options have been updated to enable/disable `autoRefresh`; it's enabled by default. Note that `autoRefresh` refreshes the entire current dashboard, so some users may wish to disable it.

Also added some explanatory comments for `useAlarmTableMenuActions`.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-470

### DIFF
--- a/src/panels/alarm-table/AlarmTableControl.tsx
+++ b/src/panels/alarm-table/AlarmTableControl.tsx
@@ -26,14 +26,36 @@ export const AlarmTableControl: React.FC<PanelProps<AlarmTableControlProps>> = (
     const { client } = useOpenNMSClient(props.data?.request?.targets?.[0]?.datasource)
     const { filteredProps, page, setPage, totalPages } = useAlarmProperties(props?.data?.series[0], props?.options?.alarmTable)
     const { table, menu, menuOpen, setMenuOpen } = useAlarmTableMenu(alarmIndexes, rowClicked, filteredProps, setState)
-    const { actions, detailsModal, setDetailsModal } = useAlarmTableMenuActions(state.indexes,
-      props?.data?.series?.[0]?.fields || [], () => setMenuOpen(false),
-      props?.options?.alarmTable?.alarmTableAdditional?.useGrafanaUser || false, client)
+
+    const { actions, detailsModal, setDetailsModal } = useAlarmTableMenuActions(
+      state.indexes,
+      props?.data?.series?.[0]?.fields || [],
+      () => setMenuOpen(false),
+      (actionName: string) => refreshPanel(actionName),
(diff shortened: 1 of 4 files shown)


src/panels/alarm-table/AlarmTableControl.tsx
```diff
@@ -26,15 +26,37 @@ export const AlarmTableControl: React.FC<PanelProps<AlarmTableControlProps>> = (
...
-    const { actions, detailsModal, setDetailsModal } = useAlarmTableMenuActions(state.indexes,
-      props?.data?.series?.[0]?.fields || [], () => setMenuOpen(false),
-      props?.options?.alarmTable?.alarmTableAdditional?.useGrafanaUser || false, client)
+
+    const { actions, detailsModal, setDetailsModal } = useAlarmTableMenuActions(
+      state.indexes,
+      props?.data?.series?.[0]?.fields || [],
+      () => setMenuOpen(false),
+      (actionName: string) => refreshPanel(actionName),
+      props?.options?.alarmTable?.alarmTableAdditional?.useGrafanaUser || false,
```
mikewrosey, line 35: Sonar would suggest using the `??` operator here, considering useGrafanaUser is a boolean. But the difference is almost negligible so just fyi.

mikewrosey (approved):
LGTM

PR #937, opened 2023-12-07
OPG-468: Add labelFormat and valueFormat to EntityDS nodeFilter

synqotik:
First part of OPG-468.

Allows specifying a `labelFormat` and `valueFormat` for the Entity data source `nodeFilter()` query.

Can be one of ['id', 'label', 'id:label', 'label:id', 'fs:fid'] for either of these.

`labelFormat` is the format for display, for example in a template variable dropdown list or in the Preview of values.

`valueFormat` is the format of the value used to interpolate variables, send to the server, etc.

Additional attributes can be specified, as before, including template variables.

For example, specifying the template variable query
```javascript
nodeFilter(labelFormat=label,valueFormat=id,location=$location)
```
would filter nodes by the location specified in the template variable `$location`, display the node label and use the node id in a query.

This also includes a bug fix so that `nodeFilter` can have a template variable specified in an attribute clause, as well as some formatting and type specifier fixes.

Still need to apply this to Performance data source, will be a separate PR.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-468

### DIFF
--- a/src/datasources/entity-ds/EntityDataSource.ts
+++ b/src/datasources/entity-ds/EntityDataSource.ts
@@ -1,31 +1,37 @@
 import { DataQueryResponse, DataSourceApi, DataSourceInstanceSettings, QueryResultMeta } from '@grafana/data'
-import { TemplateSrv, getTemplateSrv, getBackendSrv } from '@grafana/runtime'
-import { API, Model } from 'opennms'
-import { EntityTypes } from '../../constants/constants'
-import {
-    getColumns,
-    getEntityTypeFromFuncName,
-    getPropertyComparators,
-    getQueryEntityType,
-    getSearchProperties,
-    getTemplateVariable,
-    isLocationQuery,
-    isMetricMetadataQuery,
-    isSituationAttribute,
-    metricFindLocations,
-    parseFunctionInfo,
-    queryEntity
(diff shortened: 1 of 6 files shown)


mmahacek (commented):
Seems to work great on my test.  While it was confusing at first how an invalid value for either of these reverts to defaults, it's a nice QOL touch to not break it if the user puts the wrong thing in.
I might suggest adding an option for `fs:label`, but it's not a blocker for me.  The only thing holding me back from hitting approve is wanting to see an update to the docs page for Entity DS.

synqotik:
> Seems to work great on my test. While it was confusing at first how an invalid value for either of these reverts to defaults, it's a nice QOL touch to not break it if the user puts the wrong thing in. I might suggest adding an option for `fs:label`, but it's not a blocker for me. The only thing holding me back from hitting approve is wanting to see an update to the docs page for Entity DS.

I just opened this ticket for documentation: https://opennms.atlassian.net/browse/NMS-16270

I'll do that after the PR for updating the Performance data source with this feature. Also, I'll add an `fs:label` option in that PR (both for Entity and Perf DS).

mmahacek (approved):

PR #938, opened 2023-12-08
OPG-468: Refactor FunctionFormatter into individual functions

synqotik:
Refactored `function_formatter.ts` into individual functions vs. static methods in a class, this is much more "Typescript-y". Also can import just the functions you need to use. Fixed formatting, etc.

Moved `parseFunctionInfo` there since it will be reused by the Performance data source.

Also added `fs:label` specifier for both label and value.

Next PR will be to add `labelFormat` and `valueFormat` to the Performance data source `nodeFilter` query.

Refactored the `format_formatter.spec.ts` test to use `test.each` to remove a bunch of code duplication. Fix various Sonar issues. Marked one regex as `// NOSONAR` as it's not trivial to fix and has been in production for years.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-468

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -1,11 +1,22 @@
 export const EntityTypes = {
-    Alarms: 'Alarms',
-    Nodes: 'Nodes',
-    IPInterfaces: 'IP Interfaces',
-    SNMPInterfaces: 'SNMP Interfaces',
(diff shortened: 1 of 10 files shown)


mmahacek (approved):
LGTM.

PR #943, opened 2023-12-13
OPG-473: Fix Alarm Panel bug when calculating page rows

synqotik:
Fixed a bug in the Alarm Panel when calculating which rows of data to display for the given page.

Looks like we assumed `field.values` was always an `ArrayVector` having a `buffer` array field which we were accessing directly, however `field.values` is a `Vector` and might only contain `length` and `toArray()` fields. Using `toArray()` instead, and also cleaning up some code (use `slice()` instead of `splice()`, etc.).

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-473

### DIFF
--- a/src/panels/alarm-table/hooks/useAlarmProperties.ts
+++ b/src/panels/alarm-table/hooks/useAlarmProperties.ts
@@ -1,5 +1,5 @@
 import { useState, useEffect } from 'react'
-import { ArrayVector, DataFrame } from '@grafana/data'
+import { ArrayVector, DataFrame, Field } from '@grafana/data'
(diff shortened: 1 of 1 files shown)


mmahacek (approved):
This fix works for me.

synqotik:
> @synqotik Hi, recently I've been using this plugin to display data on an alarm table and stumbled upon this very issue that you solved here, so I was able to fix it by upgrading to a version of the plugin that included these changes.

Thank you, that may be a bug. I will open an issue for it and we will address it before the next release.

synqotik:
@jereparla we have a fix, will be in a release sometime this month. Thanks for reporting this!

PR #953, opened 2024-01-09
OPG-471: Perf DS allow manual override extensions, template vars in dropdowns

synqotik:
Addresses OPG-471 and OPG-384.

Adds template variables to the Performance query editor node and resource dropdowns, so they are easy to select.

Adds a new Performance data source configuration option, "Allow Manual Override Extensions". If enabled, this will display an on/off switch and an input text box toe allow users to manually enter a value for a field. For example, to enter a template variable or static value for a Node in a Performance Attribute query. Some dropdowns have validation that won't accept such values, so this allows an override mechanism.

It's likely the manual override extensions will need to be added to other fields, I wanted to get a version of this into the codebase for possible future use. It's disabled by default.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-471,  http://issues.opennms.org/browse/OPG-384


mmahacek (commented):
The inclusion of variables in the drop down somewhat negates the need to allow overrides, but both are very welcome.
With the override, it would be nice to have on both the Node and Resource fields in the first patch.  What's the timeline for adding a resource override? Before or after the .13 release?

synqotik:
> The inclusion of variables in the drop down somewhat negates the need to allow overrides, but both are very welcome. With the override, it would be nice to have on both the Node and Resource fields in the first patch. What's the timeline for adding a resource override? Before or after the .13 release?

Good point, I'll add to the Resource field as well. Also good since I think I'll refactor that code into a reusable component.

I'll have it done tomorrow, it will go into the 9.0.13 release.

RangerRick (dismissed):

synqotik:
Refactored to component, added override to Resource as well.

mmahacek (approved):

mmahacek (commented):
Something looks like it broke here.  If I have variables for the $node and $resource, when I set the $node variable, the resource drop down only shows available variables and doesn't show the options for the node's resources.  This worked in the previous commit.
Selecting a valid $resource variable DOES work, just the drop down doesn't offer them.

synqotik:
Merging this *BUT* will look into the issue above and fix in a subsequent PR.

PR #992, opened 2024-05-02
OPG-478: Alarm Table pagination fixes for Grafana 9 and 10

synqotik:
Fix some Alarm Table pagination issues. Pagination controls were in some cases displaying but not clickable. 

This was due to use dynamically adding click handlers for the context menu to a `div` that included the `Pagination` control; this moves the control outside that `div`, among other fixes.

Tested with Grafana v9.4.7, v9.5.3, v10.4.2.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-478

### DIFF
--- a/src/panels/alarm-table/AlarmTableControl.tsx
+++ b/src/panels/alarm-table/AlarmTableControl.tsx
@@ -39,6 +39,8 @@ export const AlarmTableControl: React.FC<PanelProps<AlarmTableControlProps>> = (
     const { tabActive, tabClick, resetTabs } = useAlarmTableModalTabs()
     const { alarm, goToAlarm, alarmQuery } = useAlarm(props?.data?.series, soloIndex, client)
 
+    const paginationRef = useRef<HTMLDivElement>(null)
+
     useAlarmTableRowHighlighter(state, table)
     useAlarmTableConfigDefaults(props.fieldConfig, props.onFieldConfigChange, props.options)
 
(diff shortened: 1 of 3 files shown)


christianpape (approved):
LGTM.

mershad-manesh (approved):

jereparla:
Thanks a lot for this, @synqotik.  I'm sorry for barging in this way, if there is another way you prefer me to report issues please let me know. 

We've found another bug in the alarm table panel: when "scroll" is enabled, and there are more entries in the table than fit in the panel (i.e., you *need* to scroll down in order to see all entries), if you double-click on any given entry the "Alarm Detail" window that pops up is not the correct one (does not correlate with the entry that was doubled-clicked). For example, I could click on an entry at the bottom of the table after scrolling down that has an alarm ID 80, however the Alarm Detail window that pops up is for alarm ID 10. For the entries at the top of the table that fit in the panel without having to scroll down, this doesn't happen and it works fine.

We're using Grafana 10.0.2 and the plugin at 9.0.13.

synqotik:
@jereparla thanks for the comment, and I will add a Jira issue for that bug as well, looks like we need to fix that before the next release.

We really appreciate you raising these issues! Another way to report is to use our Mattermost channel: https://chat.opennms.com/opennms/channels/helm.

synqotik:
This is *finally* fixed in OPG v12 which should be released this week.

PR: https://github.com/OpenNMS/grafana-plugin/pull/1062

PR #1047, opened 2025-09-09
Fix: TypeError in PerformanceAttribute for Performance DS (OPG-492)

sbernabe1:
Description:
This pull request fixes a TypeError that occurred when processing data from the Performance data source. The previous code was attempting to access a property on an undefined variable, which would cause the plugin to fail.

The fix ensures the variable is correctly checked for a value before being used, preventing the error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9661,9 +9661,6 @@
       }
     },
     "node_modules/follow-redirects": {
-      "version": "1.15.6",
-      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.6.tgz",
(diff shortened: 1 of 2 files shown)


synqotik:
@sbernabe1 thank you for looking into this!

Do you know what the `propertyValue` value is when this occurs? Perhaps we can update `PerformanceAttributeItemState` to have a compatible type definition. If it is `null` or `undefined` then this solution is good, but if it's a value, we should update our type definition so we do not lose data.

Also, please target branch `release-9.x` instead of `develop`. I know (and apologize) this isn't well documented.

synqotik:
> @marshallmassengill & @synqotik, I hope I am doing this right... trying to tag someone for OPG fix.

You can tag me on all OPG reviews, thanks!

PR #1048, opened 2025-09-14
Fix: Handle undefined value in performance attribute Jira-OPG-492

sbernabe1:
Description:
This pull request fixes a TypeError that occurred when processing data from the Performance data source. The previous code was attempting to access a property on an undefined variable, which would cause the plugin to fail.

The fix ensures the variable is correctly checked for a value before being used, preventing the error. 

Release-9.x 

Please Review:
@synqotik
@marshallmassengill

### DIFF
--- a/src/datasources/perf-ds/PerformanceAttribute.tsx
+++ b/src/datasources/perf-ds/PerformanceAttribute.tsx
@@ -101,12 +101,11 @@ export const PerformanceAttribute: React.FC<PerformanceAttributesProps> = ({
      * doesn't have the same resources, so clear them
      */
     const setPerformanceStateNode = async (propertyValue: unknown) => {
-      const node = propertyValue as PerformanceAttributeItemState
-
-      if (!node) {
-        return
-      }
-
+    if (!propertyValue) {
+        return;
+    }
+    const node = JSON.parse(JSON.stringify(propertyValue)) as PerformanceAttributeItemState
+    
       const resourceOptions: OnmsResourceDto[] = await loadResourcesByNode(node.id || node.label)
       const existingLabel = performanceState?.resource?.label
 
--- a/src/datasources/perf-ds/PerformanceAttribute.tsx
+++ b/src/datasources/perf-ds/PerformanceAttribute.tsx
@@ -101,11 +101,18 @@ export const PerformanceAttribute: React.FC<PerformanceAttributesProps> = ({
      * doesn't have the same resources, so clear them
      */
     const setPerformanceStateNode = async (propertyValue: unknown) => {
-    if (!propertyValue) {
-        return;
-    }
-    const node = JSON.parse(JSON.stringify(propertyValue)) as PerformanceAttributeItemState
-    
+      if (!propertyValue) {
+        return
+      }
+
+    const propertyValueAny = propertyValue as any
+
+    const node = {
+        id: String(propertyValueAny.id ?? ''),
+        label: String(propertyValueAny.label ?? '')
+    	} as PerformanceAttributeItemState	
+     
+
       const resourceOptions: OnmsResourceDto[] = await loadResourcesByNode(node.id || node.label)
       const existingLabel = performanceState?.resource?.label
 


synqotik:
@sbernabe1 @dino2gnt @marshallmassengill Thanks for finding this issue.

Could someone confirm what value for `propertyValue` is being passed to `setPerformanceStateNode` when this `TypeError` occurs?

Is it `null` or `undefined`? Then we can check `if (!propertyValue)` and return. This could be a stopgap measure just to get around this issue for now.

If it is something else, but it's not a `PerformanceAttributeItemState`, then it could mean that Grafana v12 has changed their API and somehow a different value is being passed. This conversion may be occurring elsewhere, so we would need to do some investigation.

Also if this worked before, but now the value is always `null` or `undefined` then it probably also mean Grafana changed their API and we aren't mapping the value correctly and thus losing data.

In any case we should not need to `stringify/parse` the object here, you'll get a `TypeError` anyway.

synqotik:
@sbernabe1 thanks for the info!

The issue might be that in `OnmsNode`, `id` is a `number` instead of a `string`.

Try this out. First, exit if `!propertyValue`. Then do the following to convert it to a `PerformanceAttributeItemState`.

```
const setPerformanceStateNode = async (propertyValue: unknown) => {
  if (!propertyValue) {
    return
  }
  
  const node = {
    id: String(propertyValue.id ?? ''),
    label: String(propertyValue.label ?? '')
  } as PerformanceAttributeItemState
    
    // the rest of the function. Do not need 'if (!node)` here.
}

```

sbernabe1:
@synqotik, the following is the output when trying to test the updated code provided:


    `const setPerformanceStateNode = async (propertyValue: unknown) => {
    console.log('setPerformanceStateNode, propertyValue:')
    console.dir(propertyValue)
      if (!propertyValue) {
        return
      }
        const node = {
                id: String(propertyValue.id ?? ''),
                label: String(propertyValue.label ?? '')
        } as PerformanceAttributeItemState
      //const node = JSON.parse(JSON.stringify(propertyValue)) as PerformanceAttributeItemState

      const resourceOptions: OnmsResourceDto[] = await loadResourcesByNode(node.id || node.label)
      const existingLabel = performanceState?.resource?.label

      let resource = performanceState?.resource ?? {}

      if (!isTemplateVariable(resource)) {
          const resourceOption = (existingLabel && resourceOptions && resourceOptions.filter(r => r.label === existingLabel)?.[0]) || null

          if (resourceOption?.id && resourceOption?.label) {
            resource = {
              id: resourceOption.id ?? '',
              label: resourceOption.label ?? ''
            }
          }
      }

      const state = {
        ...performanceState,
        node,
        resource
      } as PerformanceAttributeState

      setPerformanceState(state)
    }`



    
**Error's from "rpm run watch":**
 
image

sbernabe1:
@synqotik, I've made some updates to fix the error while still using the suggestions you recommended.


    const setPerformanceStateNode = async (propertyValue: unknown) => {
    console.log('setPerformanceStateNode, propertyValue:')
    console.dir(propertyValue)
      if (!propertyValue) {
        return
      }
        // Correcting Property error's.
    const nodeAsItemState = propertyValue as PerformanceAttributeItemState

    // suggested code.
    const node = {
        id: String(nodeAsItemState.id ?? ''),
        label: String(nodeAsItemState.label ?? '')
        } as PerformanceAttributeItemState
      //const node = JSON.parse(JSON.stringify(propertyValue)) as PerformanceAttributeItemState`
      
      
**Looks like this works, no need for JSON implementation:**

image

synqotik:
@sbernabe1 ahh ok, try this instead, it's just a Typescript issue, need to cast `propertyValue: unknown` to an `any` first:

```typescript
      if (!propertyValue) {
        return
      }
      
      const propertyValueAny = propertyValue as any
      
      const node = {
        id: String(propertyValueAny.id ?? ''),
        label: String(propertyValueAny.label ?? '')
      } as PerformanceAttributeItemState
 ```
 
Hopefully this works!

synqotik:
Nice! Looks good. I think it was just that the `id` is being passed in as a `number` and the `TypeError` was because we were assuming it was a `string`. Thanks for your work on this :)

synqotik:
@sbernabe1 you can go ahead and make that change and I'll approve.

synqotik (approved):
LGTM!

synqotik:
I went ahead and merged this

PR #1061, opened 2026-03-05
OPG-514: Update to Grafana 12.4.0. Add CLAUDE.md. Fix 'npm run dev'

synqotik:
Grafana just released v12.4.0, so we may as well update. Seems like a lot of changes and bug fixes.

Fixed an issue where `npm run dev`, i.e. building and running in `development` mode, was failing.

`ForkTsCheckerWebpackPlugin` in `.config/webpack.config.ts` is only created in `development` mode. Ultimately it makes calls to the `ajv` library, which were failing.

Basically, `ajv` has other dependencies like `ajv-keywords` which all have to be update in step with each other, otherwise things like undefined properties result.

We had to update `ajv` using `overrides` due to OSV security issues, and various `ajv` are used, so this can be tricky.

Used Claude Code to help with that `ajv` issue, so added a `CLAUDE.md` as well.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-514

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,127 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What This Is
+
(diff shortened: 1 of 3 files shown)


cgorantla (approved):
LGTM!

PR #1062, opened 2026-03-10
OPG-480: Fix Alarm Table Panel scrolling bug

synqotik:
Fixes an issue in the Alarm Table Panel where when you scroll the alarm items and then select one or more items, it was selecting incorrect items. So user might see the wrong alarm when opening the context menu, or when performing actions such as `Acknowledge`.

The core issue was a mismatch between the row indices, whether they started at the top of data or the visible region during scrolling, and the alarm IDs matching the rows, and mismatches between the `props` data being passed in vs. the filtered and sorted data being used in the table.

Ideally we could attach some non-visible data to each row with the alarm ID, but it was difficult to find a way to dig into the `Table` component to be able to do that. So now we enforce that the ID (alarm ID) column is always shown and we can get the alarm ID value from that (getting text out of the HTML elements).

This fix also handles the case where user goes from viewing the Panel in a dashboard to editing it and vice versa. Scrolling and selecting work correctly.

Also pulled some code from `alarm-table/hooks` up into `AlarmTableControl.tsx` so we can more clearly see the `filteredPropsRef`, `filteredPropState`, `table` and other parameters, making it a bit less convoluted and easier to understand and debug.

Lastly, the API calls for actions (acknowledge, clear, etc.) are now performed in parallel. `Promise.allSettled` is used so we make all the calls even if one fails, and then report the results to the user.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-480


cgorantla (approved):
LGTM!

PR #1111, opened 2026-08-25
OPG-521: Fix repeating panels and rows under Dashboard Scenes

synqotik:
# OPG-521: Fix repeating panels and rows under Dashboard Scenes

**Target branch:** `release-12.x`

Repeating a panel or row over a template variable produced one clone per value, but every clone queried for *all* of the variable's values — so each panel showed every instance instead of just its own.

This provides a fix which works with both Scenes and pre-Scenes. It automatically detects which is in use and handles both. So, OPG 12 should now handle repeating panels/rows in Grafana 10, 11, 12, 13, with/without `?scenes=false` in the URL and with/without `dashboardScenes = false` in the `grafana.ini`.

## Root cause

Grafana 12 renders dashboards with Scenes, which changed where a repeat's value lives.

Before Scenes, the value arrived in `DataQueryRequest.scopedVars` as `{ node: { value: '2' } }`. Under Scenes, `SceneQueryRunner` sends only:

```ts
private _scopedVars = { __sceneObject: wrapInSafeSerializableSceneObject(this) }
// later merged with __interval / __interval_ms
```

The repeat value is now a `LocalValueVariable` attached to the clone's node in the scene graph. It is reachable by passing `scopedVars` back to `templateSrv.replace()`, which routes to `sceneGraph.interpolate()` and resolves from the panel's position in the tree — but **not** via `templateSrv.getVariables()`, which always answers at dashboard scope.

All three datasources read variable values in ways that miss the local value:

| Datasource | Problem |
|---|---|
| **perf-ds** | `collectInterpolationVariables()` read `variable.current` directly. `interpolate()` then fanned every selected value into its own measurement source, in every clone. This is OPG-521 as reported. |
| **entity-ds** | The multi-value path read `variable.current` for its "All", is-multi and current-values decisions, producing an OR over every selected value. That path never consulted `scopedVars` at all, **so entity repeats were broken before Scenes too** — this is not purely a Grafana 12 regression. |
| **flow-ds** | `convertTemplateVariables()` called `replace()` with no `scopedVars`, falling back to `window.__grafanaSceneContext`, i.e. dashboard scope. |

## The fix

New `src/lib/variableResolution.ts` which answers *"what are the effective values of this variable, for this query?"*, and all three datasources route through it:

```ts
resolveVariable(templateSrv, variable, scopedVars) -> {
  name, values, texts, isLocalOverride, isAllInDashboardScope
}
```

Under Scenes it resolves via `templateSrv.replace('${name:json}', scopedVars)` and detects a repeat by resolving a second time *without* `scopedVars` and comparing. This deliberately uses only the public `TemplateSrv` interface rather than taking a dependency on `@grafana/scenes` just to test for `LocalValueVariable`.

**Repeated rows needed no separate handling.** `RowItemRepeater` attaches the same `SceneVariableSet([LocalValueVariable])` to each row clone and descendants inherit it, so a panel inside a repeated row resolves identically to a repeated panel.

### Pre-Scenes support retained

Grafana 13 removes the legacy renderer, but this branch still supports Grafana 10/11 and Grafana 12 with `?scenes=false` or `dashboardScene = false` (in `grafana.ini`). You can set `SUPPORT_NON_SCENES_GRAFANA = false` in OPG 13 and delete `resolveInLegacyScope()` to remove pre-Scenes support.

Note, flipping it to `false` will fail 4 tests (remove those tests also if removing pre-Scenes support).

### Also fixed

An "All" selection stored as the bare string `'$__all'` rather than the array `['$__all']`. Only the array form was expanded, so a single-select variable with `includeAll` leaked a literal `$__all` into the query.

## Testing

Unit tests added. Two new test doubles in `src/test/react/support/templateSrvFakes.ts` model both Scenes and pre-Scenes interpolation engines and the two ways they differ: what exists in `scopedVars`, and that `getVariables()` always returns dashboard scope. They deliberately share one formatter, because Grafana 10.4, 11.6 and 12.4 all import `formatRegistry` from `@grafana/scenes` in `formatVariableValue` — one registry serves both paths.

**Manual testing** against live Grafana 12 + OpenNMS, with a multi-value `hrStorageIndex` variable: repeated panels, repeated rows, non-repeated multi-value fan-out, the `All` selection, and the legacy path via `?scenes=false` — all tested correct. Also verified at the payload level, not just visually: each repeated clone issues a `POST /rest/measurements` with exactly one entry in `source`.

`test/dashboards/opg-521-repeat-test-dashboard.json` contains a Grafana dashboard that can be used to test this functionality. Import it via **Dashboards → New → Import**; it declares `__inputs` so it prompts for the datasource and there are no UIDs to edit. Documented under a new `## Testing` section in `DEVELOPMENT.md`.

> **Note for reviewers:** panel **B** in that dashboard is the one to look at. A repeat fix that pins the variable too aggressively breaks multi-value fan-out in *non-repeated* panels, and neither the panel count nor the request count reveals it — each clone has always had its own `SceneQueryRunner`, so the number of requests looks identical whether the code is fixed or broken. Only the length of the `source` array distinguishes them.

## Addressed in review

See comment added below rather than have this here in the description.

## Other commits in this PR

These are not part of the OPG-521 fix but were needed to get CI green.

**`Update dependencies to clear osv-scanner findings`** — `osv-scanner` reported 147 vulnerabilities across 39 packages, blocking the `plugin-validator` run in strict mode. Now there are 3 issues, none critical.

**`Point repository links at the OpenNMS-Plugins org`** — 11 references across 8 files, following the org move. Also drops the CircleCI build badge from the README, which pointed at the old org and was failing the CI link check.

## Known remaining osv-scanner findings

Both are upstream and **neither ships in `dist/module.js`**:

- **`react-router@6.30.6`** (2 findings, CVSS 6.1 / 5.1) via `@grafana/ui` → `react-router-dom-v5-compat`, whose newest release is 6.30.6 while the advisories are fixed only in react-router 7.18.0. `@grafana/ui` is a webpack external, need Grafana to fix this.
- **`fs@0.0.2`** (MAL-2025-21003, no fix available) — declared by `opennms@2.6.1`, the latest release. Tracked separately in Jira; the fix is to drop the dependency in `opennms-js`, since `fs` is a Node builtin. See [NMS-20266](https://opennms.atlassian.net/browse/NMS-20266).

## Docs

- New `## Testing` section in `DEVELOPMENT.md`.
- Corrected the transient-dependency instructions in both `CLAUDE.md` and `DEVELOPMENT.md`, which said to delete only `package-lock.json`. Reusing `node_modules` makes npm keep the installed copies, so new overrides silently do not apply — this actually bit during this work.
- `CHANGELOG.md` entry for v12.0.2. Worth noting the changelog had no entries past v9.0.14; only added the OPG-521 line rather than backfilling 10.x–12.x.

## Not included

- `src/plugin.json`'s "License & Terms" link points at `blob/master/LICENSE`. Probably ok since `origin/master` exists and latest release version will get merged back to it.
- `site.yml` and `docs/antora.yml` still say `grafana-version-tested: '12.0'`. There appears to be an `origin/docs/bump-tested-version` branch already covering this. Can update to `12.4`.

---

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/OPG-521


marshallmassengill (dismissed):
I would ship this as is.

Claude did have some other findings we may want to look at.  None should be blockers for this:

  1. variableResolution.ts:105 — parseInterpolatedJson returns [raw] for a scalar, so a JSON-quoted scalar keeps its quotes: "web-01" → ['"web-01"']. Only reachable through a JSON.stringify-everything :json registry, which is exactly what coreFormat in the fakes models — but Grafana 12.4's legacy formatVariableValue imports formatRegistry from @grafana/scenes, so that divergence no longer exists in any supported version and coreFormat's json branch is dead test code (the legacy resolver never calls :json). Either tighten to typeof parsed === 'string' ? [parsed] : [raw] (keeps 1e5 intact, unlike String(parsed)) or drop the third "way they differ" from both header comments. It matters for real once SUPPORT_NON_SCENES_GRAFANA flips to false on Grafana 10/11.
  2. variableResolution.ts:213 — isSceneInterpolationAvailable tests window.__grafanaSceneContext for truthiness, but Grafana gates its own fallback on ctx && ctx.isActive (template_srv.ts:87, 205, 283). With a truthy-but-inactive context and no __sceneObject, replace() falls through to the redux variable store — empty under Scenes — and returns ${v:json} verbatim, which the "no interpolatable value" early return reports as values: []. That path is live: PerformanceQueryEditor.tsx:171 calls collectInterpolationVariables(getTemplateSrv()) with no scopedVars, and interpolate(...)?.[0].value on the next two lines throws a TypeError when the product comes back empty (?.[0] guards the array, not element 0). Mirroring the .isActive check is a one-liner.
  3. lib_variable_resolution.spec.ts:96 — the adhoc guard is asserted with a fake whose replace echoes the target, but real Grafana returns '' for adhoc variables (formatVariableValue's isAdHoc branch), so the guard is probably dead and adhoc resolves to ['']. Worth confirming against a dashboard with an adhoc filter, since it feeds the same empty-values path as #2.
  4. ${var:text} (OPG-466) has no test anywhere in src/test/, and this PR replaced its data source: current.text → a derived texts with a silent fall-back to values on length mismatch. One entity-ds case, repeated and not, would lock it down.
  5. Cost moved from a property read to interpolation: substitute() calls resolveVariable per clause (2–3 templateSrv.replace() plus a getVariables() scan each) and again on recursion, and perf-ds resolves every dashboard variable rather than only the referenced ones. Fine at normal sizes; a per-request memo keyed on variable name would remove it if a big filter tree ever shows up.

marshallmassengill (approved):
Looks good.

PR #1113, opened 2026-08-31
OPG-526: Make the jQuery/flot dependency explicit, and related fixes

synqotik:
# OPG-526: Make the jQuery/flot dependency explicit, and related fixes

## Why

Grafana has historically loaded jQuery and flot globally at boot, and our Alarm and Flow
histogram panels relied on that. Grafana is considering removing both from the
boot path ([grafana/grafana#131346](https://github.com/grafana/grafana/pull/131346)),
which would break those panels. This branch makes the dependency explicit
so the panels keep working, and cleans up a few things found along the way.

**jQuery is not going away.** It stays available to plugins — Grafana is only
making it load on demand. So we do not need to bundle our own copy, and nothing
here changes what ships in the plugin bundle.

### 1. Declare jQuery and flot explicitly

Both panels now import what they use explicitly, and the flot modules are marked external so Grafana keeps serving them.

Also removed the `flot` and `flot-axislabels` dependencies: nothing in `src/`
ever imported them. This drops a git-URL dependency on a personal fork.

### 2. Remove jQuery from the flow-histogram legend

`setLegend()` was the last jQuery use outside the two chart calls, and it used
document-wide selectors — so two flow-histogram panels on one dashboard would
fight over the same legend. It is now plain DOM, scoped per panel instance.

The standing TODO suggested using flot's own `legend.container` option instead.
That option is broken in the flot version Grafana vendors and throws on every
render; the TODO has been replaced with the reason.

Remaining jQuery footprint is now exactly two `$.plot()` calls.

### 3. Stop requesting the invalid alarm ID `0`

Selecting a range of rows in the Alarm Table fired a request for
`api/v2/alarms/0`, which 404s. The range-selection branch used `0` to mean
"nothing selected" where the rest of the code uses `-1`. Fixed at both ends —
the assignment and the guard that let it through.

Also hardened the row ID parser, which returned `0` rather than `-1` for a blank
cell. That could previously put `0` into the selection set, so a bulk
acknowledge or clear would act on alarm `0`.

### 4. Silence the i18next Locize notice in tests

`i18next` prints a marketing notice on init, which appeared once per test suite.
It is a transitive dependency via `@grafana/i18n` and `@grafana/ui`, and is external at
runtime, so this was test-only noise. Opted out via an env var in `jest.config.js`.

The notice was removed in i18next v26, but both Grafana packages require `^25`,
so upgrading was not worth the risk for no runtime benefit.

## Testing

Unit tests added. Both panels tested and work in Grafana 9, 10, 11, 12, and 13.

## Notes

- Not addressed, and unchanged in behaviour: a background right-click leaves
  `lastClickedAlarmId` at `0`, which makes the next shift-click select a single
  row rather than extend the range. Fixing it needs the branch condition
  reworked; worth doing only if users have noticed.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-526

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,6 +2,13 @@
 // generally used by snapshots, but can affect specific tests
 process.env.TZ = 'UTC';
 
+// i18next prints a Locize marketing notice from init(), which @grafana/i18n and
+// @grafana/ui trigger on import. It is suppressed automatically when NODE_ENV is
+// 'production', but jest runs with NODE_ENV='test'. i18next removed the notice
+// entirely in v26, but @grafana/i18n and @grafana/ui both require i18next ^25.
+// https://github.com/i18next/i18next/issues/2407
+process.env.I18NEXT_NO_SUPPORT_NOTICE = '1';
+
 const { grafanaESModules, nodeModulesToTransform } = require('./.config/jest/utils');
 
 module.exports = {
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,8 +16,6 @@
         "@grafana/schema": "^12.4.0",
         "@grafana/ui": "^12.4.0",
         "async": "^3.2.3",
-        "flot": "^0.8.3",
-        "flot-axislabels": "https://github.com/j-white/flot-axislabels#master",
         "lodash.clonedeep": "^4.5.0",
         "lodash.escaperegexp": "^4.1.2",
         "opennms": "^2.6.1",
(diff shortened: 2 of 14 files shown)


marshallmassengill (approved):
No blockers but a few things you may want to address:

 - useAlarmTableSelection.ts:50 — the shift-range branch still adds ids unfiltered. Demonstrated: a range across a blank-ID row gives [101, -1, 103], and a click on an id absent from rows gives [999, undefined]. Both reach loopAction unfiltered, so Acknowledge fires doAck(-1) / doAck(undefined); allSettled keeps it non-fatal, so the cost is a bogus request and a misleading error toast. One > 0 guard in the loop closes both. (The ?? -1 on the two findIndex lines is dead.)
 - alarm_table_solo_alarm.spec.ts:42 — not.toContain(0) can't fail now that the blank row yields -1; assert toEqual([9839, -1, 9700]) or <= 0 instead. 
 - flot-modules.d.ts:3 — "importing one makes Grafana load flot and patch it onto the shared jQuery" is present tense, but #131346 is still open and Grafana main already loads flot eagerly while mapping every jquery.flot* name to { fakeDep: 1 }, so the imports are inert today. Hedge that clause; the rest of the comment, including the jQueryFlotDeps citation, is accurate.

PR #1114, opened 2026-09-01
OPG-528: Dashboard converter correctness review (v9 → v12)

synqotik:
# OPG-528: Dashboard converter correctness review (v9 → v12)

## Background

In the OPG v9 work, we created a dashboard converter tool to convert Dashboards from Grafana / OPG v8 -> v9 format. We've updated this to convert from v9 -> v12. I wanted to review the converter tool code to fix any bugs and improve it. Also test coverage had been lacking.

## Summary

An audit of `src/lib/dashboard-convert`, concentrating on the v9 → v12 conversion, against
`@grafana/schema` 12.4.9 (`dashboard.types.d.ts`). The findings are grouped A–G below, in the
order they were worked through, one commit per group.

The converter had no test coverage at all. It now has 193 new tests across five spec files,
including golden snapshots of two full dashboards, idempotency checks and structural invariants.

Most findings were silent: the conversion produced a dashboard that looked fine and imported
without error, but had lost a field, pinned a value the source never set, or written a value in a
shape Grafana would not resolve. Two of the more consequential problems were found by the new
tests rather than by reading the code, and are noted where they came up.

---

## A. Correctness bugs

Nine confirmed defects, each reproduced before being fixed.

**A.1 — `mapRequires` mutated the input and never wrote the output.**
`obj = { ...r }` copied the entry *before* `r['version'] = 12` mutated the original, so the version
bump was a no-op and the caller's `dashboardV9` was corrupted. Now assigns to the copy.

**A.2 — Annotation target limit read from a property that does not exist.**
`convertToInt(obj.target.number)` — `AnnotationTarget` has `limit`, not `number`. Every annotation,
including the built-in "Annotations & Alerts", was given `limit: 0`. Now reads `limit`, defaulting
to Grafana's 100.

**A.3 — `AnnotationPanelFilter.ids` mapped to strings.**
The schema types it `Array<number>`; the mapper used `convertToString`, so panel-scoped annotation
filters matched nothing. Now uses `convertToInt`.

**A.4 — `convertToInt` returned `NaN` instead of the default.**
The fallthrough `return val` returned the `NaN` it had just computed, which `JSON.stringify` writes
as `null`. This affected every numeric field in the converter and is shared with the perf, entity
and flow datasources. The default is now held in a separate const the parse cannot clobber.

**A.5 — `isEnumValueOfType` accepted TypeScript reverse-mapping names.**
`Object.values(VariableHide)` yields `['dontHide', …, 0, 1, 2, 3]`, so `hide: "hideVariable"` passed
the guard and then became `NaN` via A.4. Numeric enums now match on numbers only.

**A.6 — `source.editable` was never mapped.**
Hard-coded `true`, so a read-only dashboard became editable on conversion. Now mapped.

**A.7 — `Source = 9, Target = 9` returned an error.**
Both are offered in the UI, but the combination fell through to "Invalid targetVersion". Added
`normalizeV9.ts`: the dashboard is already v9, so it does only the housekeeping that makes the Json
safe to re-import (strip `uid`, apply title, set/increment version). Converting 10/11/12 down to 9
is a downgrade and now says so explicitly.

**A.8 — `unhideAllQueries` and `convertGraphToTimeSeries` were dead on the v9 → v12 path.**
Both were passed in and never read. `ConvertOptions` is now threaded through the panel mappers so
they apply, including to panels nested in a collapsed row. The graph conversion matters more here
than on the v8 path: the Angular `graph` panel was removed in Grafana 11 and cannot render under
v12.

**A.9 — `JSON.parse` was unguarded.**
Malformed Json threw out of `dashboardConvert` into the panel's `onClick`. Now returns an
`isError` response, as the v8 path already did.

## B. Silent data loss

**B.1 — Mappers dropped every field they did not enumerate.**
`mapAnnotationQuery`, `mapDashboardLink`, `mapVariableOption` and `mapVariableModel` built an
allowlist and discarded the rest. They now spread the source and overwrite only what they
normalize, which recovers `filters` / `baseFilters` / `defaultKeys` on adhoc variables,
`defaultValue` on groupby variables, `regexApplyTo` and `valuesFormat`, `properties` on a variable
option, `placement` on annotations and links, and datasource-specific annotation query fields —
both on the annotation and inside its `target`, which for a non-Grafana datasource *is* the query.

**B.2 — `mapVariableType` was missing `'switch'`.**
New in v12; a switch variable was silently rewritten to `'query'`.

**B.3 — Target-level datasource refs were not normalized.**
Fixed as part of C, which it depended on: routing targets through `mapDataSourceRef` beforehand
would have turned a working reference into `null`.

## C. Datasource references

`mapDataSourceRef` put a template variable string into the ref's `type` with an empty `uid`, which
Grafana cannot resolve, and dropped a plain datasource name string to `null` entirely.

Both now map to the `uid`, which is how Grafana resolves a `DataSourceRef` and what its own schema
migration (`migrateDatasourceNameToRef`) does. No special-casing of variables versus names is
needed, so the legacy `[[datasource]]` syntax is covered for free, and the empty `apiVersion` /
`uid` placeholders are gone. Targets are now normalized the same way as panels (B.3).

Both now map to the `uid`, which is how Grafana resolves a `DataSourceRef` and what its own schema
migration (`migrateDatasourceNameToRef`) does, and the empty `apiVersion` / `uid` placeholders are
gone. Targets are normalized the same way as panels (B.3), and a variable reference is emitted in
the braced spelling whichever of the three it arrived in (R3.1).

## D. Injected defaults

`createEmptyV12Dashboard` seeded values Grafana does not default to. Because every field is copied
from that seed when the source omits it, converting a dashboard silently changed how it behaved.
Absent now maps to absent, with fallbacks keyed off `defaultDashboard` / `defaultGridPos` /
`defaultAnnotationQuery` so they track the installed `@grafana/schema` rather than drifting.

**D.1 — Dashboard level.** `schemaVersion: 0` made Grafana replay every migration from the start
(a source value is now kept as-is); `liveNow: true`, `preload: true` and `refresh: '1m'` are no
longer invented; `timezone` defaults to `'browser'`, not `'utc'`; `weekStart`, `description`,
`gnetId` and `revision` are no longer written out empty; `id` is nulled and `uid` omitted, matching
Grafana's own export for sharing externally.

**D.2 — Sub-object level.** An annotation with no explicit `enable` is no longer switched off; a
missing `gridPos` gets Grafana's 12×9 rather than a zero-sized panel, and `static` is only written
if the source set it; `quick_ranges` and `refresh_intervals` are omitted rather than emitted empty,
which would blank the refresh picker; `staticOptions: []` is no longer stamped on every variable;
a variable option keeps `selected` only if present, and an empty multi-value selection stays `[]`
instead of collapsing to `''`.

## E. Panel `pluginVersion`

`pluginVersion: '12'` was stamped on every panel. Grafana uses that field to decide which of a
plugin's migrations to run, so writing it onto a core `timeseries` or a third-party panel
misrepresents which migrations have been applied — and `'12'` was not a real version for our own
panels either.

Panels typed `opennms-*` now get the version of that panel's own plugin, looked up by plugin id
in the `pluginVersions` module. Everything else keeps whatever the source had, and gets none
invented if it had none.

## F. Typing

Every mapper ended in an `as Panel` / `as VariableModel` style assertion, which suppressed exactly
the conformance errors that would have caught A.2 and A.3.

Deleting the assertions would have achieved nothing: where a mapper spreads the untyped source,
TypeScript widens the whole literal to `any`, so a declared type is not checked either. The fields
we normalize are now built as a separate typed object and merged over the source on the way out,
which keeps the passthrough behaviour while making our own output checked.

**Found while doing this:** a `.map` over an `any` returns another `any`, so array fields were
still unchecked. The new `mapArray` helper gives them a real element type — and fixes the latent
bug it exposed: `.map(convertToString)` passes the array index as the function's second argument,
which is its `defaultValue`, so `['a', null, 'c']` came back as `['a', 1, 'c']`. This affected
annotation target tags, dashboard link tags and `refresh_intervals`.

*Correction to the review: `mapRowPanel` does not collide row ids on 0 — the assignment is guarded
and an absent id stays absent.*

## G. Tests

The converter had no tests. Added fixtures under `src/test/react/support/fixtures`: a Helm v8 and an OPG v9 dashboard, written to exercise the shapes the mappers have to handle.

**G.1 — Golden.** Snapshots of the whole converted dashboard for v9 → v12, v8 → v12 with options
off and on, and v8 → v9. Any unintended change to the output is now a reviewable diff.

**G.2 — Idempotency.** Converting an already converted dashboard must not keep changing it. Holds
for the v9, v8 and v9-normalize paths. The version increment is intentional and is asserted as such
rather than worked around.

**G.3 — Invariants.** No top-level key dropped, no panel lost at any nesting depth, no `NaN` left
in the output. The `NaN` walk runs over the object rather than the Json, because `JSON.stringify`
writes `NaN` as `null`, where it cannot be told apart from the nulls the schema does allow
(`Threshold.value`, a default datasource, the id). A small self-check keeps the walk honest.

**Found while doing this:** `addVariationsToMap` called `replaceAll` with a non-global `RegExp`,
which throws a `TypeError`. It is reached from `parseTemplating` for any `datasource` type template
variable pointing at an OpenNMS datasource — the ordinary shape of a Helm dashboard — and nothing
catches it between there and the panel's `onClick`, so the Convert button simply did nothing. This
means **v8 → v9 conversion has been broken for the common case.**

---

## Additional fixes

**Datasource template variable in the bare string form.** `updateTargetDatasource` is documented as
retaining the uid when the datasource is a template variable, but the check read
`source.datasource.uid`, which is undefined for the bare string `'$datasource'` rather than
`{ uid: '$datasource' }`. Those panels and targets were rewritten to a concrete uid taken from
whichever Grafana ran the conversion — defeating the variable and producing a uid that will not
exist on the importing instance. Both forms are now handled; a string with no variable sigil is a
datasource name and still resolves to the concrete uid.

**`__requires` version as a semver string.** Grafana reads `__requires[].version` as a minimum
version constraint, so the bare number `12` was the wrong shape. OpenNMS datasource and panel
entries now get the version of the plugin the entry names, from the same source as E.

**R1 / R6 — `__requires` and `pluginVersion` read the app's version.** A new
`pluginVersions` module keys the nested versions by plugin id, resolving a legacy Helm id to its
current equivalent; an id we do not recognise is now left alone rather than given a guess.

**Null sub-objects threw out of `dashboardConvert`.** `isDefined` is `!== undefined`, so a
null passes it and the guarded call then dereferences. New `isDefinedObject` guards every call that dereferences.

*Legacy Helm datasource plugin id strings became a dangling uid.** A v8 dashboard can name
its datasource by plugin id rather than by variable. `getSourceDatasourceInfo` now recognises a bare
OpenNMS plugin id, legacy or current, and reports it as a non-variable so it resolves to the
installed datasource; a new `annotations` module runs annotation queries through the same rewrite
as panels and targets.

**`normalizeV9` ignored two options the panel always offers.** It read only
`incrementDashboardVersion`, but `DashboardConvertPanelControl` renders "Unhide all queries" and
"Convert Graph to Timeseries Panels" regardless of target version, so Source 9 → Target 9 with both
on applied neither. `applyPanelOptions` now walks the panels, including those in a collapsed row.

**`delete panel.seriesOverride` was missing its trailing `s`.**
`convertLegacyGraphToTimeSeriesPanel` read `seriesOverrides` into the fieldConfig overrides and
then left the dead legacy array on the panel.

---

## Addressing Reviewer's Comments

Twelve further issues. Each was reproduced against the code before being fixed; several were
broader than reported, and three were mine to un-do rather than extend.

**R2.1 — `addVariationsToMap` did not register the `[[name]]` spelling.** All three variable
spellings are valid, so a panel using `[[DS_ONMS]]` was reported as non-OpenNMS and the v8 *query*
conversion never ran — the same dashboard converted correctly through `$DS_ONMS` and `${DS_ONMS}`
but emitted a raw v8 target for the bracket form. The C claim that the bracket syntax was "covered
for free" was true of `mapDataSourceRef` only.

**R2.2 — Three crashes in the v8 graph converter.** `source.yaxes?.[0].label` was missing its
second `?.` (the sibling line has it), and `convertFieldConfigOverrides` read `o.alias` off
unguarded entries. `yaxes: []`, `yaxes: [null]` and `seriesOverrides: [null]` each threw.
Threading the options through the panel mappers for A.8 made these reachable from the v9 paths,
where nothing catches them — the same escaped-`TypeError` symptom as R2 in round 1.

**R2.3 — `__elements` was dropped.** Verified against Grafana v12.4.0:
`manage-dashboards/import/utils/inputs.ts` reads it to resolve library panel inputs, so dropping it
breaks any dashboard using one. The spread-and-overwrite rewrite had been applied to sub-objects
but not to the top level, and the "no top-level key dropped" invariant passed only because neither
fixture carried the key.

**R2.4 — Unrecognised values were erased, not preserved.** The `{ ...obj, ...typed }` merge wrote
`undefined` for anything the mapper could not interpret, and that `undefined` overwrote the source.
`{hide:'2', refresh:'1', sort:'3'}` came out with none of them. Two fixes: `mapEnumValue` converts
before checking so numeric strings are accepted, and `withoutUndefined` drops unset keys before
merging so a value we genuinely cannot interpret is left alone.

**R2.5 — `convertToInt(null, d)` returned 0.** `Number(null) === 0`, so null was parsed rather than
treated as absent. `schemaVersion: null` became 0 against a schema default of 42 — the exact
migration-replay problem D set out to fix. `schemaVersion: 'abc'` became 0 too, because that call
passed no default.

**R2.6 — `mapFieldConfigSource` used `isDefined`,** where the round-1 null-guard commit established
`isDefinedObject`, so `{defaults: null}` passed through against a non-optional field.

**R2.7 — `convertLegacyGraphToTimeSeriesPanel` hard-coded `pluginVersion: '9.4.7'`,** which is
exactly what E established is not ours to write on a panel that is not one of our plugins.

**R2.8 — `pluginVersion` churned back to base's value.** Accepted: the nested plugins declare
`"12"` and webpack templates `%VERSION%` only into the root `plugin.json`, so the looked-up value
cannot differ from the string base hard-coded. E's behavioural fix is the *scoping* — see the note
in section E.

**R2.9 — The `__requires` rationale rested on behaviour Grafana 12 does not have.** Confirmed
against v12.4.0: `exporters.ts` writes it, `api.go` and `migrate.go` delete it unread. The change
is cosmetic and the "Not addressed" bullet that claimed a stale entry suppresses an import warning
was wrong; both are corrected above.

**R2.10 — Four mappers were still allowlists.** `mapDashboardSnapshot`, `mapTimePickerConfig`,
`mapGridPos` and `mapTimeOption` dropped unknown fields, and the snapshot mapper padded a partial
object with empty strings and zeroes. All four now spread and overwrite.

**R2.11 — Empty arrays were still invented.** A panel with no links, targets or transformations,
and a variable with no options, were given empty arrays — the D rule applied everywhere except
here.

**R2.12 — `id` differed by path.** Nulled on v9 → v12, kept on v8 → v9 and v9 → v9, all three of
which already dropped the `uid`. Now consistent.

**Removed rather than extended: the annotation datasource rewrite.** Round 1 asked for it; this
round asked which dashboards produce the case it handles. The answer is none. Every OPG datasource
declares `"annotations": false` and implements no `annotationQuery`, and the Helm datasources it
converts from declared the same, so an annotation has never been able to be backed by one. It was
also incomplete — it rewrote `annotation.datasource` but not the target query shape — so a
hypothetical hand-edited dashboard would have come out looking converted but broken. The reachable
half of that round-1 fix stays: `getSourceDatasourceInfo` recognising a bare legacy plugin id on a
panel or target.

## Review round 3

**R3.1 — A retained `[[name]]` datasource ref does not resolve after import.**
`getTemplateVariableUid` echoed the source spelling into the emitted `uid`, and Grafana 12 rejects
the bracket form in both places it would have to be resolved. Verified against v12.4.0:

| | `$name` | `${name}` | `[[name]]` |
|---|---|---|---|
| Import substitution of an `__inputs` value (`dash_template_evaluator.go`, `varRegex` is `(\$\{.+?\})`) | ✗ | ✓ | ✗ |
| `getInstanceSettings` (`datasource_srv.ts`, interpolates only when `nameOrUid[0] === '$'`) | ✓ | ✓ | ✗ |
| `get()` (interpolates unconditionally) | ✓ | ✓ | ✓ |

So a `templating.list`-backed `[[name]]` ref runs its queries but shows as missing in the
datasource picker, and an `__inputs`-backed one fails outright. `$name` misses the import
substitution too — that predates this PR, but `${name}` is the only spelling that survives
everything, so all three now normalize to it.

Safe for a ref with no matching `__inputs` entry: the evaluator returns an unmatched `${...}`
unchanged rather than blanking it, and the runtime interpolates it. Applied on the v9 → v12 path
as well, where `mapDataSourceRef` had the identical exposure. Registering `[[name]]` in the
datasource map stays — without it the v8 *query* is never converted (R2.1).

---

## Not addressed

- `__requires` still passes the `{ "type": "grafana", "version": "8.5.0" }` entry through
  untouched. An earlier draft of this section claimed that suppresses an import warning; it does
  not — Grafana 12.4 deletes `__requires` without reading it, so the stale entry is inert.
- `mapDashboardLink` still writes `url: ''` when the source has none. Harmless, and the only
  optional field left following the D pattern.
- The v8 converter's legacy plugin id regex (`^opennms-helm-([^-]+)-datasource$`) predates this
  work. The current Helm repo declares the ids without the `helm-` infix, so the legacy form may
  come from an older Helm release; it was left as found.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/OPG-528


marshallmassengill (changes requested):
@marshallmassengill all addressed, see the Description, `Addressing Reviewer's Comments` section.

marshallmassengill (changes requested):
One last blocker... or at least I think it is one:

Retained [[name]] datasource refs do not resolve after import

getTemplateVariableUid (convert-from-v8/utils.ts:118-125) echoes the source spelling into the emitted uid, so a v8 panel using [[DS_ONMS]] converts to { "type": "opennms-performance-datasource", "uid":"[[DS_ONMS]]" }.

Grafana 12 rejects that in both places it would have to be resolved:

- Import substitution matches ${NAME} only: varRegex = regexp.MustCompile("(\\$\\{.+?\\})") in dashboardimport/utils/dash_template_evaluator.go:14, with inputs registered as e.variables["${"+inputName+"}"].
- getInstanceSettings interpolates only behind if (nameOrUid[0] === '$') (plugins/datasource_srv.ts:115); a bracketed uid falls to the literal map lookup at :140 and returns undefined. get() at :174 does interpolate, so a templating.list-backed ref still runs its queries while the datasource picker shows it missing. An __inputs-backed ref, which is what the new test builds, fails on both paths.

Bare `$DS_ONMS` misses the import substitution too. That predates this PR; ${NAME} is the only form that survives both.

Possible Fix: return '${' + rawName + '}' for all three spellings. The new dashboard_convert_v8.spec.ts cases assert the current uids and need updating with it. Adding [[name]] to the datasource map is correct and should stay, since without it the v8 query is never converted.

marshallmassengill (approved):
LGTM!

PR #1115, opened 2026-09-02
OPG-527: Fix the RPM build and add packaging tests

synqotik:
# OPG-527: Fix the RPM build and add packaging tests

Audited the `makerpm.js` RPM-generation script for correctness and found that the RPM it produced was unusable. Fixing that turned up related problems in `makedeb.js`, `makezip.js` and the build itself, all of
which are addressed here. Also removes our dependency on `speculate` and adds a test suite for
package generation.

## Main issue

**The RPM has been installing the plugin somewhere Grafana never looks.** PR #1054 replaced
`specit` with `bbc/speculate`. `specit` honored `spec.specTemplate` and `spec.installDir` from
`package.json`; **`speculate` 6.x does not** — `speculate/lib/spec.js` hardcodes its own bundled
template and silently ignores both. `src/rpm/spec.mustache` was therefore dead config, and what
actually shipped was speculate's systemd-Node-service spec.

So the package installed to the wrong prefix, created a system user, enabled a systemd unit that
does not exist, required `nodejs`, shipped a stray `/var/log` directory and a stray `.service`
file, and owned its files as the wrong user.

### RPM generation

`speculate` was dropped in favor of rendering `src/rpm/spec.mustache` ourselves.

`makerpm.js` is now a thin CLI shim over testable modules:

| Module | Responsibility |
| --- | --- |
| `scripts/rpm/spec.js` | Renders `src/rpm/spec.mustache` |
| `scripts/rpm/archive.js` | Packs the `dist` tree into the source tarball rpmbuild consumes |
| `scripts/rpm/build.js` | Lays out an rpmbuild tree, runs rpmbuild, publishes the artifact |
| `scripts/packageVersion.js` | Derives version and release from `package.json` (shared with `makedeb.js`) |
| `scripts/distContents.js` | What never belongs in a distributable (shared by all three scripts and the webpack config) |
| `scripts/deb/maintainer.js`, `scripts/deb/metadata.js` | Debian maintainer identity, `control` and `changelog` |
| `scripts/webpack/excludeFromCopy.js` | Applies the exclusion list to `CopyWebpackPlugin` |

Dependencies: `-speculate`, `+mustache`, `+tar-fs`.

The resulting RPM installs 81 files, all under `/var/lib/grafana/plugins/opennms-opennms-app`,
with no scriptlets and `grafana >= 12.0.0` as its only real dependency — matching what the DEB
has always done.

### Bugs fixed along the way

- **`clean()` targeted `cwd`, not `dist`** — but `generate()` wrote into `dist`, so a leftover
  `dist/SPECS` from an earlier run made the next build fail with `EEXIST`.
- **`rpmbuild`'s exit status was never checked** — only `spawnSync`'s `error`. A failed build went
  unnoticed until the artifact copy blew up with a confusing message. This one matters below.
- **The built RPM's filename was reconstructed rather than discovered**, ignoring `%{?dist}` and
  `spec.noarch`, so a successful build could be reported as a failure.
- **`isDebug = true` was committed.** Now `MAKERPM_DEBUG=1`.
- **`makedeb.js` released everything as 0** — `if (version.indexOf('-SNAPSHOT'))` is truthy for
  `-1` too. Only affected local builds, since CI always passes `--release` explicitly.
- **A read failure mid-archive left a truncated `.tar.gz`** that looked complete, because `.pipe()`
  forwards neither errors nor teardown.
- **`makezip.js` shipped the jest fixtures** and did not check `zip`'s exit status either.

### The DEB signing failure

Checking `rpmbuild`'s exit status has a counterpart in `makedeb.js`, and turning that check on
revealed a failure that had been silently discarded for as long as it has existed:

```
gpg: skipped "Benjamin Reed <ranger@opennms.org>": No secret key
dpkg-buildpackage: error: failed to sign .dsc file
dpkg-buildpackage exited with status 25
```

`dpkg-buildpackage` builds the `.deb`, then tries to clear-sign the `.dsc` as the changelog's
maintainer identity — a key we do not hold. Because the `.deb` was still produced and the exit
code ignored, the job went green. It now passes `-us -uc`: CI signs the built `.deb` separately
with the OpenNMS release key, so signing here was both redundant and using the wrong identity.

The former maintainer's name is also no longer used. `debian/control` is generated from
`src/debian/control.mustache` alongside the changelog so both name one identity — the OpenNMS
build account, overridable via the standard `DEBFULLNAME`/`DEBEMAIL` variables.

### Grafana version consistency

The DEB required `grafana (>= 9.0)` while `package.json` and `plugin.json`'s `grafanaDependency`
both said `12.0.0`. The mismatch ran in the awkward direction: a Debian user on Grafana 9 could
install the package cleanly and then get a plugin Grafana refuses to load. `Depends` is now
derived from the same `spec.requires` the RPM uses, translating RPM operator spelling to dpkg's.

The requirements doc was updated to match (it still claimed the plugin "may work with older
versions" and "most likely Grafana 10 and 11", and gave `9.x` as the recommended pairing), along
with the stale Horizon 31+ / Meridian 2023+ line.

### MANIFEST.txt vs package contents

Fixed by excluding files at the source (`webpack.config.ts` applies `scripts/distContents.js` as a `CopyWebpackPlugin` ignore).

> Note for reviewers: `scripts/webpack/excludeFromCopy.js` reaches into the `CopyWebpackPlugin`
> instance's `patterns` property, which is **not documented API** — `.config/` is scaffolded and
> `webpack-merge` concatenates plugin arrays rather than reconfiguring the existing plugin. It
> throws a build-time error naming itself if it cannot find the plugin, because silently not
> excluding would reintroduce this bug.

## Tests and Validation

`src/test/packaging/` adds **80 tests** across 7 suites, including golden snapshots, full `rpmbuild`, real DEB built and inspected, etc.

## Behavior changes worth noting

1. **The DEB now refuses to install on Grafana < 12.** Intended, and it matches what Grafana
   enforces at plugin load, but it is a real change for anyone who had installed it onto an
   older Grafana.
2. **The ZIP no longer contains the jest fixtures.** It previously shipped them.
3. **`makerpm.js`, `makedeb.js` and `makezip.js` now fail when the tool they invoke fails.**
   Previously a non-zero exit from `rpmbuild`, `dpkg-buildpackage` or `zip` was ignored. Expect
   these jobs to start reporting problems that were previously invisible.
4. `debian/control` is now generated. `src/debian/control` became `src/debian/control.mustache`.

## Follow-ups (not in this PR)

- The same treatment — analysis, extraction, tests — for `makedeb.js` and `makezip.js` beyond the
  fixes made here.
- Consolidating the packaging layout, moving the `makeXXX.js` scripts into `scripts/` along the rest of the files
- The pre-existing `node_modules/opennms` typecheck failures.
- `debian/control` still declares `Standards-Version: 3.7.3` (Debian Policy 3.7.3 dates from
  2007; current is 4.7.x) and `Build-Depends: debhelper (>= 9)` against a `debian/compat` of 10.
  Whether `Section: contrib/net` is still the right archive area is also worth a look.

Also fixed the PR template with correct Jira and CI URLs.

# External References

* JIRA (Issue Tracker): https://opennms.atlassian.net/browse/OPG-527
* Continuous Integration: [GitHub Actions](https://github.com/OpenNMS-Plugins/grafana-plugin/actions)


marshallmassengill (approved):
No blockers from me.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -25,5 +25,5 @@ Thanks for taking time to contribute!
 
 # External References
 
-* JIRA (Issue Tracker): http://issues.opennms.org/browse/${JIRA-ISSUE-NUMBER}
-* Continuous Integration: [CircleCI](https://circleci.com/gh/OpenNMS/grafana-plugin)
+* Jira (Issue Tracker): https://opennms.atlassian.net/browse/${JIRA-ISSUE-NUMBER}
+* Continuous Integration: [GitHub Actions](https://github.com/OpenNMS-Plugins/grafana-plugin/actions)
```
mershad-manesh, line 29: Did we switch to Github Actions for CI/CD?

PR #1116, opened 2026-09-03
OPG-530: Move the packaging scripts and their inputs under scripts/

synqotik:
Consolidates everything that builds a distributable into one place, one directory per artifact type. Doing this before continuing work on auditing/fixing `make-deb.js` and `make-zip.js`.

| Directory | Files | Original Location |
| --- | --- | --- |
| scripts/rpm/ | make-rpm.js, spec.mustache | (from makerpm.js, src/rpm/) |
| scripts/deb/ | make-deb.js, debian/ | (from makedeb.js, src/debian/) |
| scripts/zip/ | make-zip.js | (from makezip.js) |

The templates in particular do not belong under src/: webpack sweeps src/ into dist with broad copy globs and `npm run sign` then attests everything in dist into a signed MANIFEST.txt, so a packaging input with a matching extension would ship in the plugin and break the signature.

Adds `npm run package:rpm|deb|zip` so .circleci/config.yml invokes the scripts by name rather than hardcoding their paths.

Beyond the paths, the three entry points now resolve dist/ and artifacts/ from __dirname rather than process.cwd(). That was only safe while they sat at the repository root; they no longer do.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-530


marshallmassengill (approved):
LGTM!

PR #1117, opened 2026-09-04
OPG-529: Audit the deb and zip builders and make them testable

synqotik:
Gives `make-deb.js` and `make-zip.js` the treatment `make-rpm.js` got under OPG-527 / PR #1115: a thin entry point over modules that take paths as arguments, so the pipeline can be exercised against a fixture dist.

Bugs the audit found, all now covered by tests:

- Both scripts only cleaned up on the success path, so any failure left a full copy of dist behind. The deb built inside artifacts/ itself, which is where CI collects build artifacts, and the zip staged into a project-root `tmp/` that it then removed wholesale despite not owning it. Both now stage under the system temp directory and clean up in a `finally`.
- `dpkg-buildpackage` writes a `.dsc`, `.changes`, `.buildinfo` and source tarball beside the `.deb`. Building in `artifacts/` meant CI stored all of them; only the `.deb` is signed and published. The deb is now copied out and the rest discarded with the build tree.
- Neither confirmed its artifact existed. The deb never even reported where it went. Both now report the path, and the deb asserts exactly one was built rather than trusting the exit status.
- A trailing `.catch` reported every failure in the chain as "Copy failed", including template and write errors. A missing dist now says to run `npm run build`, the way `make-rpm.js` already did.
- `make-deb.js` mutated the required `package.json` module object with the resolved version and release. Nothing read either field: `renderControl` takes name/description/spec.requires and `renderChangelog` takes version and release as explicit arguments.
- Errors went to `stdout`. They now go to `stderr`, matching `make-rpm.js`.

Adds `MAKEDEB_DEBUG` and `MAKEZIP_DEBUG` to match `MAKERPM_DEBUG`; both scripts were previously unconditionally verbose.

Extracts s`scripts/stageDist.js` (dist staging plus the actionable missing-dist error, shared by deb and zip) and `scripts/artifacts.js` (`copyToArtifacts`, moved out of `scripts/rpm/build.js` now that all three builders publish the same way).

The zip keeps naming its artifact with the raw `package.json` version. The rpm and deb split version from release because their formats need it for sort order; a zip does not, and the suffix distinguishes a snapshot from a release. CI consumes the zip through a glob, so nothing depends on the name.

Verified: full suite (445 pass, 4 skipped), lint, typecheck, build, and real end-to-end builds of all three artifacts.

Note: `.circleci/config.yml` does not run the jest suite at all, so these tests are a local check rather than a CI gate.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-529

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -192,7 +192,11 @@ Monitor https://community.grafana.com/t/build-a-panel-plugin-error/100984/3 for 
   `/usr/lib` with a bogus service unit instead of into the Grafana plugin directory
 - Only includes `dist/` in the RPM (not `node_modules`), minus the entries in
   `scripts/distContents.js`
-- Set `MAKERPM_DEBUG=1` for CircleCI debugging
+- Set `MAKERPM_DEBUG=1` for CircleCI debugging (`MAKEDEB_DEBUG` / `MAKEZIP_DEBUG` for the
+  other two)
+- The deb and zip follow the same shape: a thin entry point over a testable
+  `scripts/<type>/build.js`, staging via `scripts/stageDist.js` and publishing via
+  `scripts/artifacts.js`. Neither builds inside `artifacts/`
 
 ## Support Matrix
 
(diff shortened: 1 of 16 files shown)


mershad-manesh (approved):
Looks good.

PR #1118, opened 2026-09-08
OPG-517: Link the Alarm Table detail dialog to the OpenNMS alarm page

synqotik:
The "Full Details" link in the Alarm Table panel's alarm detail dialog never worked.

`opennms-js` builds `Alarm.detailsPage` from the datasource url, which is Grafana's proxy path ('/api/datasources/proxy/uid/<uid>'), so the link pointed back at Grafana rather than at OpenNMS. The url in the HTTP settings is no help either: it is one the Grafana server must reach, and is typically a Docker `host.docker.internal` address that a browser cannot reach.

Adds an optional "OpenNMS Base URL" setting to the Entity datasource, which is the base url of an OpenNMS instance as the user's browser reaches it, and used that as the base for links. The OpenNMS-relative portion is recovered by stripping the datasource url prefix, so the alarm page path stays owned by `opennms-js`.

Grafana permits this: `TextLink`'s external branch emits a plain anchor and deliberately skips `locationUtil.stripBaseFromUrl`, `textUtil.sanitizeUrl` leaves absolute http(s) urls untouched, and the plugin validator's link analyzers (`safelinks`, `brokenlinks`) only cover `plugin.json` and documentation urls.

If no base url is configured, the link is shown as an OpenNMS-relative path in plain text, not as a link, since a relative href would resolve against Grafana's own origin and lead nowhere. Enabling the setting without entering a url marks the field invalid rather than silently degrading to that state.

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-517

### DIFF
--- a/docs/modules/datasources/pages/entity_datasource.adoc
+++ b/docs/modules/datasources/pages/entity_datasource.adoc
@@ -10,7 +10,9 @@ For an example of the Entities datasource in use, see <<panel_configuration:dyna
 In addition to the standard HTTP settings, the Entities datasource has an *Additional Options* section with an *OpenNMS Base URL* setting.
 
 Some panels can link directly to a page in the OpenNMS web UI, such as the *Full Details* link in the Alarm Table panel's alarm detail dialog.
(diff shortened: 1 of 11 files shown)


marshallmassengill (approved):
LGTM!

PR #1119, opened 2026-09-08
OPG-500: move ci to newer base images

synqotik:
# OPG-500: Move CI to newer base images

Moves the CI packaging and docs jobs onto currently maintained base images, then fixes what
that exposed: the docs build has been failing, and xref validation has never actually gated
anything.

Five commits:

- Move CI to current base images and fix the docs build
- Make xref validation strict and fix a broken cross-page reference
- Validate that xref anchors exist, not just the pages they live on
- Install node dependencies in `build-docs`
- Address review findings in the docs validators and the RPM job

## 1. The base images

`opennms/build-env` has had no RPM- or Debian-family push since **November 2023**. Its newest
RHEL-family tag is CentOS 8 with **Node 16**; its newest Debian tag is bullseye with **Node
18**. Both are well below this repo's `engines` (`>=22 <25`), which is what was breaking the
deb build.

| executor | was | now |
|---|---|---|
| `build-executor-centos` → `rpm-build-executor` | `opennms/build-env:11.0.14_9-3.8.7-b9528` (CentOS 8, Node 16) | `rockylinux/rockylinux:9` |
| `build-executor-debian` → `deb-build-executor` | `opennms/build-env:debian-jdk11-b10453` (Debian 11, Node 18) | `node:22-trixie` (Debian 13) |
| `node-executor` | `cimg/node:22.22` | `cimg/node:22.23.2` |
| `docs-executor` | `opennms/antora:3.1.4-b10433` | **removed** — see §2 |

`make-tarball` and `make-zip` moved to `node-executor`. They need only node, `tar` and `zip`,
all present in `cimg/node`, so there was no reason to install node into Rocky for them. That
also stops `package:zip` running under Node 16.

Three things needed handling that aren't obvious from the diff:

- **`sign-packages/install-deb-dependencies` breaks on modern Debian.** 
  `make-deb` now runs a plain `apt-get update` first (it needs `debhelper` anyway); the orb
  passes `-o APT::Get::List-Cleanup=0`, so those lists survive into its own call.

- **Rocky 9 carries no `nodejs`**, so `nodejs npm rpm-build` are installed explicitly
  (`rpm-sign` comes from the signing orb). `tar`, which the spec's `%setup` shells out to,
  needs no listing — it is in the base image *and* a declared dependency of `rpm-build`.

- The `vault.centos.org` repo rewrite in `make-rpm` is gone — it was a workaround for CentOS
  8's dead mirrors and is meaningless on Rocky.

## 2. The docs build was failing

`build-docs` has been dying with:

```
"msg":"Generator not found or failed to load. Try installing the '@antora/site-generator' package."
"message":"diagChan.tracingChannel is not a function"  at node_modules/pino/lib/tools.js:32
```
`docs-executor` has now been removed and `build-docs` runs on `node-executor` using this repo's own `@antora` devDependencies, bumped
`^3.1.14` → **`^3.2.0`** (the current `latest`; everything above it is `3.2.0-rc.*`).

`@antora/xref-validator` was the one thing genuinely holding us to the old image — it isn't on
npmjs.org, only a GitLab tarball. 

`npm run validate-xrefs` now installs it itself into a gitignored `.antora-tools/`.

The docs UI bundle now comes from `OpenNMS/antora-ui-opennms` **v3.1.1**, matching
`antora-playbook-local.yml` in the main OpenNMS repo. The old `opennms-forge` bundle is a
different repo whose newest release is v3.1.0 from 2022. To be clear, the bundle is only
presentation assets — it has **no** bearing on the Node/Antora problem above; swapping it
alone reproduces the identical crash.

## 3. Xref validation never gated anything

Antora's default `failure_level` is **`fatal`**, so a broken xref logged at `error` was
*reported* and the command still **exited 0**. The "Validate Xrefs in docs" step has been
decorative all along.

Both `validate-xrefs` and `docs` now pass `--log-failure-level=warn`, which covers warn and
above — missing images and includes too, not just xrefs.

That isn't sufficient on its own. Asciidoctor reports a **dangling internal reference** —
`<<some-anchor>>` where that anchor isn't on the page — at **`info`**, and `failure_level`
accepts only `warn|error|fatal|none`, so that class can never fail on Antora's exit code.
`scripts/docs/validateXrefs.js` wraps the validator, runs it at
`--log-level=info --log-format=json`, and fails on any reference message whatever level it was
logged at.

**That gap is how a real bug survived:** `getting_started/importing.adoc` linked to
`<<upgrade-dashboards>>`, an anchor that lives on `installation/upgrading.adoc`. It rendered as
`href="#upgrade-dashboards"` on a page with no such anchor — a link to nowhere. Now a proper
`xref:installation:upgrading.adoc#upgrade-dashboards[...]`.

## 4. Anchors are now validated too

Antora validates the *page* half of an xref target and **never** the `#anchor` half, so
`xref:installation:upgrading.adoc#typo[]` resolves the page, renders a link to nowhere and
reports nothing at any log level. `scripts/docs/validateAnchors.js` closes that, running as a
third `build-docs` step because it reads the site `npm run docs` builds.

One deliberate limitation: **root-relative hrefs are skipped.** Doing it properly needs
`site.url`, which this script deliberately doesn't read; Antora emits these only for UI assets
and the 404 page, never with a fragment. Leaving as is for now.

## 5. Updated `opennms-js` to `2.7.0`

Updated the `opennms` dependency to `2.7.0`. This removed some high/critical security vulnerabilities, most
noticeably with the removal of `fs`.

There's still a few `typecheck` errors in `opennms-js`, pretty minor ones, that weren't caught in time of the `2.7.0` release. We'll fix them in a subsequent release.

## Pre-existing issues, deliberately not fixed here

- `npm run typecheck` fails with 7 errors, down from 11. These are all in `opennms-js`. We'll eventually fix these there and have a new build.
- `osv-scanner` reports 2 medium vulnerabilities. The `opennms-js` upgrade to `2.7.0` removed some high/critical ones.

----------

# External References

* JIRA (Issue Tracker): http://issues.opennms.org/browse/OPG-500

### DIFF
--- a/scripts/docs/validateAnchors.js
+++ b/scripts/docs/validateAnchors.js
@@ -1,0 +1,233 @@
+#!/usr/bin/env node
+
+'use strict';
+
+// Checks that every in-page anchor the built docs link to actually exists.
+//
+// Antora validates the *page* half of an xref target and never the `#anchor`
+// half, so `xref:installation:upgrading.adoc#typo[]` resolves the page, renders a
+// link to nowhere and reports nothing -- at any log level. That is the last silent
+// class of broken reference, and scripts/docs/validateXrefs.js cannot see it.
+//
+// This works on the generated HTML rather than the AsciiDoc source on purpose.
+// Anchors come from auto-generated section ids (subject to idprefix/idseparator),
+// `[[x]]`, `[#x]`, block ids and discrete headings; deriving that set from source
+// means reimplementing Asciidoctor, and getting it wrong reports links that are in
+// fact perfectly good. The rendered `id` attributes are ground truth.
+//
+// Run it after `npm run docs`, which is what produces the site it reads.
+
+const fs = require('fs');
+const path = require('path');
+
+const PROJECT_DIR = path.resolve(__dirname, '..', '..');
+const DEFAULT_SITE_DIR = path.join(PROJECT_DIR, 'public');
+
+// The UI bundle's assets. No pages, and its ids are not ours to police.
+const UI_ASSET_DIR = '_';
+
+// Anything with a scheme (http:, https:, mailto:, tel:, javascript:) or
+// protocol-relative is not a link into the built site.
+const HAS_SCHEME = /^([a-z][a-z0-9+.-]*:|\/\/)/i;
+
+// Asciidoctor generates these itself, in matched pairs; they are not authored.
+const GENERATED_FRAGMENT = /^_footnote(?:def|ref)_\d+$/;
+
+function collectIds(html) {
+  const ids = new Set();
+
+  for (const match of html.matchAll(/\s(?:id|name)="([^"]*)"/g)) {
+    if (match[1] !== '') {
+      ids.add(match[1]);
+    }
+  }
+
+  return ids;
+}
+
+function extractAnchorLinks(html) {
+  const links = [];
+
+  for (const match of html.matchAll(/\shref="([^"]*)"/g)) {
+    if (match[1].includes('#')) {
+      links.push(match[1]);
+    }
+  }
+
+  return links;
+}
+
+// Returns the file the fragment must live in and the fragment itself, or null when
+// the link is not ours to check.
+function resolveTarget(pageFile, href) {
+  if (HAS_SCHEME.test(href)) {
+    return null;
+  }
+
+  const hashAt = href.indexOf('#');
+  const relative = href.slice(0, hashAt);
+  const raw = href.slice(hashAt + 1);
+
+  // `href="#"` is a chrome affordance (dropdown toggles), not a reference.
+  if (raw === '') {
+    return null;
+  }
+
+  let fragment = raw;
+
+  try {
+    fragment = decodeURIComponent(raw);
+  } catch {
+    // A fragment that is not valid percent-encoding is still worth checking as
+    // written, so fall through with the raw value rather than skipping the link.
+  }
+
+  if (GENERATED_FRAGMENT.test(fragment)) {
+    return null;
+  }
+
+  let file;
+
+  if (relative === '') {
+    file = pageFile;
+  } else if (relative.endsWith('/')) {
+    file = path.resolve(path.dirname(pageFile), relative, 'index.html');
+  } else {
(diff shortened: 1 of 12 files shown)


marshallmassengill (changes requested):
One hurdle I think:

npm test does not run from a clean checkout of this branch

  - 00d9635 left the lockfile out of sync, so npm ci fails there outright.
  - 920c9f2 and 5987353 fix that, but strip resolved from 1372 of 1377 entries and integrity from 1374 (base: 0 and 2), so npm ci verifies no tarball. They also drop 48 platform binaries, leaving darwin-arm64 alone for @swc/core-* and @parcel/watcher-*. That is what breaks the binding above, and a plain npm install repairs it, which is what points at the lockfile.
  - With the binding supplied, jest 30.5.1 still cannot resolve <rootDir>/jest-setup.js on this tree. Pinning to 30.4.2 clears it and both new spec files pass, 38 tests. I could not isolate this one: the base tree with 30.5.1 force-installed runs all 335 tests, it is not the dropped jest-pnp-resolver/walker/makeerror/tmpl, and .config/ is byte-identical between the two trees.

Regenerating the lockfile on a clean checkout against the public registry, then running npm test, should fix all of those.

scripts/docs/validateAnchors.js
```diff
@@ -0,0 +1,256 @@
...
+// `[[x]]`, `[#x]`, block ids and discrete headings; deriving that set from source
+// means reimplementing Asciidoctor, and getting it wrong reports links that are in
+// fact perfectly good. The rendered `id` attributes are ground truth.
+//
+// Run it after `npm run docs`, which is what produces the site it reads.
+
+const fs = require('fs');
+const path = require('path');
+
+const PROJECT_DIR = path.resolve(__dirname, '..', '..');
```
mershad-manesh: Can we use `scripts/paths.js` here?

synqotik: @mershad-manesh yes, good catch. Should now be fixed.

marshallmassengill (approved):
LGTM!

PR #1120, opened 2026-09-12
OPG-532: Run the unit tests in CI

synqotik:
# OPG-532: Run the unit tests in CI

## Summary

CircleCI never ran the unit tests. The `build-and-sign` step was named `Run Build, Tests, and Sign`, but its body only ran `npm run build`, `npm run sign` and `release-lint.sh` — so a failing suite could not turn the pipeline red. This adds a `verify` job that runs the tests (and ESLint), and wires it into the workflow so nothing is packaged or signed from a tree with a failing suite.

This work also exposed a real latent bug in the deb builder's toolchain detection. That fix is included.

### A new `verify` job

Runs `npm run test:ci` and `npm run lint`, and requires only `pre-build` — so it runs **in parallel** with `build` and `build-docs` rather than queueing behind them.

Like `build-docs`, the packaging specs build their own fixture tree under the system temp directory, and everything else is a plain unit test.

The four `make-*` jobs now require both `build` and `verify`. `create-merge-branch` already requires those jobs, so it is covered transitively — `build-and-sign`'s other call site does not need a second test run.

### ESLint now runs in CI

Previously, ESLint did not run in CI at all. `ESLintPlugin` and `ForkTsCheckerWebpackPlugin` live in the `env.development` branch of the scaffolded webpack config, so the production build loaded neither, and the "lint" inside `build-and-sign` is OpenNMS's `release-lint.sh` (a release/version sanity check), not ESLint.

Lint runs after the test-result upload, and both upload steps include `when: always`, so the test report is always made.

### Test reporting

`jest-junit` writes `test-results/jest/results.xml` so CircleCI's TESTS tab shows per-test failures and timing. It's gated on `process.env.CI`, so a local `npm test` keeps its usual output.

This replaces the `store_test_results` / `store_artifacts` pair in `build` that pointed at `coverage/` — a directory that never existed, since jest was never asked for coverage and the tests were never run.

### Deb and Rpm toolchain detection (bugs found by running the CI build)

The first run of `verify` failed three tests in `deb_build.spec.ts` with `dpkg-buildpackage exited with status 25 / fakeroot not found`.

We checked for `dpkg-buildpackage` but none of `build-essential`, `debhelper`, `fakeroot` or `rpm / rpmbuild`, which was causing builds not to finish.

`fakeroot` is required **only when not running as root**. It is the gain-root command, and `dpkg-buildpackage` needs one only when it is not already root — which `deb-build-executor` (`node:22-trixie`) is. Verified in that image: as root with no fakeroot on PATH, `dpkg-buildpackage` exits 0 and produces the `.deb`, and the full deb suite runs and passes rather than skipping.

Both `lookup` and `asRoot` are injectable, so the partial-toolchain and root cases are covered without depending on what the host has or who it runs as — and one case calls the real `which.sync` default so the production path is exercised too.

`verify` installs `build-essential debhelper fakeroot rpm`. The first three are the list `make-deb` installs; `rpm` provides `rpmbuild`, which Ubuntu otherwise lacks.

Without these, 17 end-to-end tests would have silently skipped — a small replay of the exact problem this PR exists to fix. The RPM block in particular had **never** run in CI.

The deb end-to-end tests now pass an explicit `120000` timeout like their rpm counterparts, instead of relying on jest's `5s` default.

## Notes for reviewers

**Why `rpm` from Ubuntu rather than a job on `rpm-build-executor`.** A dedicated job would test against el9's rpm 4.16.1.3 instead of Ubuntu's 4.18.2. These specs cover our spec template and build script — package naming, metadata from `package.json`, install paths, file ownership, `Requires:`, absence of scriptlets, fixture exclusion, cleanup — none of which turn on the rpm version. A dedicated job would also have to require `build` rather than `pre-build`, since it needs `node_modules` from the workspace, so it could not run in parallel the way `verify` does. Both options were measured before choosing; the comment in `config.yml` records when to revisit.

**`npm run typecheck` is deliberately not included.** It fails due to 7 errors, every one of them inside `node_modules/opennms/src` (i.e. `opennms-js`), and none in our own `src`. We will eventually fix `opennms-js` and update OPG.

**`@grafana/plugin-validator` and the signed manifest are unaffected.** Nothing new lands in `dist`; `test-results/` is at the repo root and gitignored.

## External References

* Jira (Issue Tracker): https://opennms.atlassian.net/browse/OPG-532

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,7 @@ commands:
   build-and-sign:
     steps:
       - run:
-          name: Run Build, Tests, and Sign
+          name: Run Build and Sign
           command: |
             export PATH="./node_modules/.bin:${PATH}"
             export NODE_OPTIONS=--openssl-legacy-provider
@@ -167,14 +167,38 @@ jobs:
       - validate-schema
       - build-and-sign
       - save-node-cache
-      - store_test_results:
-          path: coverage
-      - store_artifacts:
-          path: coverage
+      # This job used to store_test_results and store_artifacts from coverage/, but it
+      # never ran the tests and jest was never asked for coverage, so the directory did
+      # not exist. The unit tests and their JUnit output live in the `verify` job now.
       - persist_to_workspace:
           root: ~/
           paths:
             - project
+
+  verify:
+    # The unit tests do not need dist: the packaging specs build their own fixture tree
+    # under the system temp directory, and the rest are plain unit tests. So this runs in
+    # parallel with `build` off `pre-build` rather than queueing behind it, the same way
+    # `build-docs` does, and costs no wall-clock on the critical path.
+    executor: node-executor
+    resource_class: medium
+    steps:
+      - attach_workspace:
+          at: ~/
+      # `build` owns the cache write; this job only restores, as `build-docs` does.
+      - install-node-dependencies
+      - run:
+          name: Run unit tests
+          command: npm run test:ci
+      # Stored before the lint step because CircleCI skips the remaining steps once one
+      # fails: uploading here means a lint failure cannot swallow the test report.
+      - store_test_results:
+          path: test-results
+      - store_artifacts:
+          path: test-results
+      - run:
+          name: Run ESLint
+          command: npm run lint
 
   build-docs:
     # Antora comes from this repo's own @antora devDependencies, not from an image.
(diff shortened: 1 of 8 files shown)


mershad-manesh (approved):
The changes looks good to me

